PR #1, opened 2019-02-08
[help.keyman.com] Lexical model schema updates

mcdurdin:
Schema documentation for the .model_info file format, and updates to the package kmp.json metadata information.

Note that only lexical-models compiler at present is capable of generating or consuming the additional fields in kmp.json.


developer/cloud/model_info/1.0/index.php
```diff
@@ -0,0 +1,385 @@
...
+<td>yes</td>
+<td>yes</td>
+<td></td>
+<td>
+  <p><b>string</b>, the possible values for <b>license</b> vary by location in the repository:</p>
+  <ul>
+    <li><b>release/</b> and <b>experimental/</b> models must be <code>"mit"</code>.</li>
+  </ul>
+  <p>This must always be specified, because this must be an explicit decision by the model author 
+    for legal reasons.</p>
```
eddieantonio: "You must make a choice, but there is only one choice" 😂

developer/cloud/model_info/1.0/index.php
```diff
@@ -0,0 +1,385 @@
...
+
+<tr>
+<td>packageIncludes</td>
+<td></td>
+<td>yes</td>
+<td>yes</td>
+<td>
+  <p><b>array</b>, of <b>strings</b>. This flags whether a package includes:</p>
+  <ul>
+    <li><code>"fonts"</code> - one or more fonts</li>
```
eddieantonio: I think the only field allowed here is "documentation", as per the API: https://github.com/keymanapp/api.keyman.com/pull/1/files#diff-759cc5b00ab0861a406375de94017782R23

developer/cloud/model_info/1.0/index.php
```diff
@@ -0,0 +1,385 @@
...
+<td></td>
+<td></td>
+<td>
+  <p><b>object</b>, with each property an id of a lexical model and corresponding value is a <b>ModelRelatedInfo</b>.</p>
+</td>
+</tr>
+</tbody>
+</table>
+
+<h3>ModelLinkInfo Object</h3>
```
eddieantonio: I'm not 100% clear what ModelLinkInfo objects are for. May you please clarify?

developer/cloud/model_info/1.0/index.php
mcdurdin: It's like all the best political systems.

developer/cloud/model_info/1.0/index.php
mcdurdin: Good catch! Sadly ... the API/schema is wrong ... :( It should be `'fonts'`, because the model of including a separate documentation file in the package is kinda deprecated, as `welcome.htm` is the root documentation file, and links to other files can be made from that. Will fix the schema.

developer/cloud/model_info/1.0/index.php
mcdurdin: These are for external URLs, e.g. https://mymodel.example.com/ for "Model Home Page" etc.

eddieantonio: Okay!

mcdurdin:
I'm kinda ready to merge these, only I am wondering if we need to split versioning of kmp.json schema?

mcdurdin:
Blocked by https://github.com/keymanapp/help.keyman.com/pull/2

Issue #20, opened 2019-03-05
[help.keyman.com] Keyman for Linux command line should be documented here


mcdurdin:
https://help.keyman.com/products/linux/11.0/ currently only contains very basic 'install keyboard' GUI documentation. This is not sufficient; the command line tools should also be documented (the man pages, plus an overview). See https://manpages.debian.org/unstable/keyman/ for sample content (which should of course be generated from the source man pages at release time).

darcywong00:
Add @mcdurdin review comment:
> The man pages should be generated from the keymanapp/keyman repo -- as the master source? https://github.com/keymanapp/keyman/tree/master/linux/keyman-config

glasseyes:
> Add @mcdurdin review comment:
> 
> > The man pages should be generated from the keymanapp/keyman repo -- as the master source? https://github.com/keymanapp/keyman/tree/master/linux/keyman-config

Note that the man pages are generated during the build/reposetup process so generating content from a clean repo won't work.

Autogenerating from what's on Debian also has potential to get out of sync if anything changes on master while Debian sticks to stable.

ermshiperete:
The current version documents the command line tools under Reference: https://help.keyman.com/products/linux/current-version/reference/, so I think we can close this issue.

darcywong00:
Transferring to the main repo as the "Getting Started" docs are now configured there.

PR #22, opened 2019-03-06
Update the Linux documentation

darcywong00:
For #20, incorporate the man pages from https://manpages.debian.org/unstable/keyman/

* Moved installing-keyboard.php as a guides
* Update common questions from keyman.com


products/linux/11.0/web.config
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mcdurdin: Is this file needed given we have web.config in the parent folder which does the same things?

darcywong00: I lifted it from `developer\engine\web\11.0`.

It seems like it's needed? Without it, I couldn't browse to /products/linux/11.0/common

mcdurdin: Okay, it's no big deal anyway :)

PR #28, opened 2019-06-19
Fix erroneous example

eddieantonio:
`index()` must match the `any()` at the third position; not the second.


mcdurdin:
Thank you :)

PR #30, opened 2019-06-24
Update documentation on support for if/set/save/reset for macOS

mcdurdin:
See keymanapp/keyman#1832

### DIFF
--- a/developer/language/reference/if.php
+++ b/developer/language/reference/if.php
@@ -73,7 +73,7 @@ if(opt2 = '1') + any(letterGroup1) > index(outputGroup1,2)
 
 <p>The <code>if()</code> statement was introduced in Keyman 8.0.</p>
 <p>Keyman 9.0: Added support for variable system stores.</p>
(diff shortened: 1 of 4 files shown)


developer/language/reference/if.php
```diff
@@ -73,6 +73,7 @@
 
 <p>The <code>if()</code> statement was introduced in Keyman 8.0.</p>
 <p>Keyman 9.0: Added support for variable system stores.</p>
+<p>Keyman for macOS 11.0.221 added support for `if()`, `set()`, `save()` and `reset()` for variable stores.</p>
```
darcywong00:
```suggestion
<p>Keyman for macOS 11.0.221 added support for <code>if()</code>, <code>set()</code>, <code>save()</code> and <code>reset()</code> for variable stores.</p>
```

I don't think the site styles Markdown ticks, so shouldn't we stick with `<code>` tags?

mcdurdin: Ugh, this is why I shouldn't work while jet lagged! :) Good catch!

PR #33, opened 2019-08-19
Move additional files from tavultesoft.com

mcdurdin:
There will be a corresponding PR from tavultesoft.com shortly.


developer/models/index.php
```diff
@@ -0,0 +1,14 @@
...
+There are two different readmes available for a model in the repository:
+
+•  README.md in the root of the model folder.
+• readme.htm in the package file.
+
+These will often overlap but have different target audiences.
+
+README.md is a fairly well understood document visible in GH browsing, targeted at developers.
+
+readme.htm will be shown if the package .kmp is installed manually (as opposed to automatically through the cloud install, which is silent) and introduces the model to the user so they know they are installing the appropriate file.
```
darcywong00: I didn't recognize this file from the old site. Is it new content?

mcdurdin: It shouldn't have been here, sorry!

PR #38, opened 2019-09-12
Baseline iOS 12.0 help seeding from 11.0

jahorton:
This PR, in its present form, serves only as a baseline copy of 11.0's content, but with 12.0 links.  It also fixes a minor broken point or two from the old content.

Note that NOTHING has been updated within this content; it's a simple 'seeding', so I'm likely to treat this branch as 'staging' once approved while everything gets updated in 'child' branches.


products/iphone-and-ipad/12.0/installing-custom-keyboards.php
```diff
@@ -0,0 +1,47 @@
...
+  Step 3) Click the top right "Install" button.</br/>
+  The keyboard from the keyboard package is successfully installed!<br/>
+  <img src="<?= cdn('img/app/12.0/dist-kmp-success-i.png')?>">
+  <br/>
+  All the keyboards in the package are installed as a group. In this example, the package only has the "GFF Amharic 7"
+  keyboard, so it becomes the active keyboard<br/>
+  <img src="<?= cdn('img/app/12.0/dist-install1-i.png')?>">
+  <br/>
+
+  To learn how to create a custom installable keyboard, <a href="http://help.keyman.com/developer/11.0/guides/distribute/">click here</a>.
```
darcywong00: You can link to https://help.keyman.com/developer/12.0/guides/distribute/

PR #39, opened 2019-09-12
Adds 'embed' mode to page templates for use with app help

jahorton:
This PR adds logic that allows the help.keyman.com section for iOS to be used for in-app help instead of (the outdated) http://keyman.com/iphone-and-ipad/app/.  Rather than maintain the same content in multiple locations, why not simply reuse the content in one location?  This gives the added benefit of splitting in-app help content across multiple accessible pages.

To access this for testing, simply append `?embed=ios` to any page under the iOS 12.0 help section, such as http://help.keyman.com.local/products/iphone-and-ipad/12.0/index.php?embed=ios, assuming your local environment is set up appropriately.  Feel free to then replace the `?embed=ios` part with `?embed=none` (due to session management) and compare.

Note that this change also will greatly facilitate resolution of https://github.com/keymanapp/keyman/issues/1943, providing a nice, versioned target for info page linkage.

---

Finally, the Android help page version has been slightly redone to use the same `?embed` parameter (as `?embed=android`) for better in-app embedding and may also take a `&formfactor` query parameter.  I'm currently using it to allow choosing whether "phone" or "tablet" is displayed first, but it could be used towards additional ends as well.  Test that via http://help.keyman.com.local/products/android/12.0/?embed=android&device=phone, again assuming your environment is set up appropriately.  The app would naturally know which form factor it's on, allowing the appropriate images to show up by default once this parameter is utilized.

Personally, I'd refactor the Android page to write the main content once within a callable PHP function.  The only thing that varies?  The image URLs, which could then be selected by a conditional upon the `$formFactor` parameter.  Such a rework would be beyond the scope of this PR, though.

Note that `&formfactor` will function perfectly fine as `?formfactor` if `?embed` goes unused, making it possible to drop the tab system entirely if we prefer that instead.  (It defaults to the current `'tablet'` if not set, and a simple link could swap to the `?formfactor=phone` version.)  Either way, that's beyond the scope of this PR; just throwing out ideas for discussion toward future improvements.


products/android/12.0/index.php
```diff
@@ -1,27 +1,61 @@
 <?php
+$embedded = isset($_GET["embed"]) ? $_GET["embed"] : false;
```
darcywong00: Booleans in PHP can be sneaky.
These two appear the same to me.
http://help.keyman.com.local/products/android/12.0/index.php?embed=true
http://help.keyman.com.local/products/android/12.0/index.php?embed=false

vs
http://help.keyman.com.local/products/android/12.0/index.php

I suspect `$embedded` is getting strings `'true'` and `'false'`.

darcywong00: Maybe this?
```
$embedded = isset($_GET["embed"]) ? filter_var($_GET["embed"], FILTER_VALIDATE_BOOLEAN) : false;
```

_includes/includes/template.php
```diff
@@ -56,27 +56,33 @@ function head($args=[]){
     }else{
         $js = array(cdn('js/kmlive.js'));
     }
-    if(isset($args['showMenu'])){
-        $menu = $args['showMenu'];
-    }else{
-        $menu = true;
-    }
+
+    $embed = isset($args['embedded']) ? $args['embedded'] : false;
```
darcywong00: I wonder if we should follow the pattern in session.php for keyman.com
(selected lines below)
```
  if(isset($_REQUEST['embed'])) {
    $embed = $_REQUEST['embed'];
    if(!in_array($embed, ['none','windows','macos','linux'])) {
      $embed = 'none';
    }
  ...
  }

  ...
  $embed_win = $embed == 'windows';
  $embed_mac = $embed == 'macos';
  $embed_linux = $embed == 'linux';
```

Then, other pages use `global $embed_win` as needed.
We could expand to use `$global_android` or `$global_ios`

_includes/includes/no-menu.php
```diff
@@ -2,14 +2,29 @@
  if(!isset($device)){
     $device = '';
  }
+
+ // If we're in embedded mode, let's disable the default help.keyman.com
+ // hyperlink at the top; it could easily be used by accident,
+ // and embedding "web views" often have no back button.
+ if(!isset($embed)) {
+   $embed = false;
```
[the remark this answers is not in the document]
jahorton: Yeah... they're two separate embed variables.  This one is triggered through template.php as a function parameter, rather than the processed, platform-specific `embed`.  I should probably use a different name for the function parameter form... just need to think of a good one.

_includes/includes/template.php
jahorton: @mcdurdin So... rename session.php to embed.php?  That's literally what the page is designed to do.

This `$embed` (I know, should have used a distinct name) is separate.  This is the definition of the template-processing version of `$embed`, which is boolean instead of string-based.  It's set by each platform's pages - so iOS help feeds this code's function with `$args['embedded'] = $embed_ios`, for example.

`template.php` doesn't automatically know _which_ $embed___` variable is correct for a page; the pages need to set that.  Once they do, this `$embed` signals that the general "embedded" variation of the template should be used.

_includes/includes/no-menu.php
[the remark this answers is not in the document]
jahorton: An example of the link between them:

```
head([
    'title' => 'Keyman for Android Help',
    'css' => ['template.css', 'app-info-a.css']
    'embedded' => $embed_android  // This is used to initialize template.php's $embed to true. 
]);
```

Cross reference with [this comment](https://github.com/keymanapp/help.keyman.com/pull/39/files/155752b92beed7cc53c9d1c8bfd3b95ee816000d#r325449546).

jahorton:
> However, I am reluctant to approve this PR because we are adding a stack of maintainability complexity to the documentation in the merge. The biggest drawback is that this makes it very difficult for l10n in the future as translators can work with basic HTML but embedded PHP will probably throw them.

Oh - I suppose we could just have translators work directly with HTML, but I'd figured we'd be looking into one of those localization programs / content managers / whatever that facilitates string replacement instead.

> We should be able to use session variables safely to avoid having to annotate links with the embed and formfactor variables. The comments in existing code about early versions of IE not honouring sessions are many years old and are in any case not applicable to mobile devices (yay!). That should clean up things somewhat.

:+1:

> I would like to see this PR reworked to use a high-level class name inserted to control the embedding mode, for example `<body class='embed-ios formfactor-tablet'>` or `<div class=...>`. Then we can control the display of content based on the embedding mode with CSS and should not need any PHP embedded AFAICT?

I might need a couple CSS pointers for that; I'm no expert there.  Know any good references for this technique?

> One concern here is that this makes the build process dependent on the availability and design of help.keyman.com. I would prefer to be drawing the content from the help.keyman.com repository and avoiding the website dependency altogether.

I'd prefer to draw from `help.keyman.com.local`, but that'd require some interesting BA setup.  I didn't want to make that too complex, so I kinda did go lazier by basing on the live help.keyman.com instead.

_includes/includes/template.php
```diff
@@ -117,7 +121,7 @@ function write_breadcrumbs(){
   
   require_once('index-content.php');
   
-  function begin_main($toc, $index){
+  function begin_main($toc, $index, $crumbs){
```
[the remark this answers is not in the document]
jahorton: It's only referenced within template.php.  It's a helper function to the same file's `head()` function that _is_ used throughout the site.

I've also tested various places within the site; no issues seem to result.

products/android/12.0/index.php
```diff
@@ -1,27 +1,61 @@
...
+  <h2>Getting Started</h2>
+  <?php
+}
+?>
 
-<p><a href='../version-history'>Version history</a></p>
+<?php
+if(!$embedded) {
+  ?>
+  <p><a href='../version-history'>Version history</a></p>
```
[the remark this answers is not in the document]
jahorton: > This should wait until 13.0; it is introducing significant complexity into the build process too late in the beta cycle. Let's turn it into an issue.
> 
> One concern here is that this makes the build process dependent on the availability and design of help.keyman.com. I would prefer to be drawing the content from the help.keyman.com repository and avoiding the website dependency altogether.
> 
> But if you do decide to scrape, use `curl` instead of `wget` and we get it with Git Bash.

Would you be opposed to going ahead and writing the script, but not making it a part of the build process?  It's still useful, even as a manual call for updating - and that would allow us to verify success and make any needed final tweaks if/as appropriate for now.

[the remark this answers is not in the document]
mcdurdin: > Would you be opposed to going ahead and writing the script, but not making it a part of the build process? It's still useful, even as a manual call for updating - and that would allow us to verify success and make any needed final tweaks if/as appropriate for now.

Sounds like an appropriate level of automation; go ahead and write the script (and document it in the relevant readme in the source repo)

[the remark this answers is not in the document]
jahorton: Also, I also would prefer to draw from `help.keyman.com.local` or something similarly robust against help.keyman.com availability, but I figured that'd require some 'interesting' BA setup we wouldn't want to do just yet.  I didn't want to make that too complex, so I kinda did go lazier by basing the script on the live help.keyman.com instead.

_includes/includes/template.php
jahorton: I've gone ahead and disambiguated the two variables; hopefully things are much clearer now.

jahorton:
Also, re:

> A final thought. There is no real reason why we need to use the (somewhat ugly, imho) tabbed format for the documentation on tablets, but, even if we continue to do so, we should still be able to control it with CSS and use PHP only to insert a class name in a high-level element.

I completely agree; that page is full of WET documentation.  That said, I feel the change is most appropriately handled within its own PR; I didn't want to dive too deep into the Android page here.

Of course, the decisions made here will factor into how the Android page gets modified, so the concerns are still quite valid for this PR's work.

products/android/12.0/index.php
```diff
@@ -1,27 +1,55 @@
...
+  $tabletClass = "selected-tab";
+} else {
+  $mobileClass = "selected-tab";
+  $tabletClass = "";
+}
+?>
+
 <div id="android-tab-header">
-  <div id="android-mobile-header">
+  <div class="<?=$mobileClass?>" id="android-mobile-header">
```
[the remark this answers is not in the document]
jahorton: This particular section of code is designed to preserve existing interactions with [kmlive.js](https://github.com/keymanapp/help.keyman.com/blob/master/cdn/dev/js/kmlive.js#L77). kmlive.js itself is untouched (so far) by this PR.

That said... upon reviewing the code in kmlive.js at the moment, we may be able to simply rely on the js to automatically select the right tab for now, rather than use PHP to have the app set it directly.  (I think KMW's phone vs tablet detection follows the same window.width rule.)

products/android/12.0/index.php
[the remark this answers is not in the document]
jahorton: OK.  Been investigating possible use of `cURL` here.  Unfortunately, `cURL` requires going one file at a time; it's far less tailored to website mirroring.  (We'd have to iteratively specify all files to download.)

`wget` makes this far simpler, more maintainable, and less fragile; it's also designed for easy use in the Git Bash... so long as the command is available.  (It was first developed for Linux.)

_includes/includes/template.php
```diff
@@ -116,8 +124,30 @@ function write_breadcrumbs(){
   }
   
   require_once('index-content.php');
+
+  // This function uses globals defined in the embed/ folder include files, applying
+  // classes useful for content filtering between online and in-app help with embed.css.
+  function build_page_class($embeddable, $do_embed) {
+    global $formFactorClass;
```
darcywong00: Is this still used?

products/android/12.0/index.php
```diff
@@ -1,20 +1,23 @@
...
+<h2 class="content-online">Keyman for Android</h2>
 
-<p><a href='../version-history'>Version history</a></p>
+<p class="content-online"><a href='../version-history'>Version history</a></p>
 
 <!-- Content below this copied into the Keyman Android info.html -->
 
 <div id="android-tab-header">
-  <div id="android-mobile-header">
+  <div" id="android-mobile-header">
```
darcywong00:
```suggestion
  <div id="android-mobile-header">
```

_includes/includes/template.php
jahorton: Yes.

darcywong00: >     $outerClass = build_page_class($embeddable, $do_embed);

I'm not seeing `$formFactorClass`

jahorton: Oh, the global.  Right, that part did get removed; good catch.

_includes/includes/template.php
```diff
@@ -130,13 +160,22 @@ function begin_main($toc, $index){
...
+
+    $outerClass = build_page_class($embeddable, $do_embed);
+    if(!empty($outerClass)) {
+      $outerClass = " class='$outerClass'";
+    }
+
     $html = <<<END
 <div class="main$tocClass">
-  <div id="section2">
+  <div id="section2"$outerClass>
```
mcdurdin: :+1:

_includes/includes/no-menu.php
```diff
@@ -2,14 +2,29 @@
...
-            <div id="no-menu-spacer"></div>
-            <img id="header-bottom" src="<?php echo cdn("img/headerbar.png"); ?>" alt='Header bottom' />
-            <a id="help" href="http://keyman.com"><p id="keyman-help">Keyman.com</p><img src="<?php echo cdn("img/helpIcon.png"); ?>"></a>
-        </div>
-        <div id="phone-header-spacer"></div>
\ No newline at end of file
+  <div id="container">
+    <div class="header">
+      <div id="show-phone-menu-spacer"></div>
+      <?php if(!$embed_template) { ?>
```
mcdurdin: So this PHP interleaving is done here because the embedding control class is further down in the document, right? I am reasonably comfortable with that because it's in the template code so maintenance is not a big issue.

PR #45, opened 2019-09-17
Allow users to zoom in on pages on mobile browsers

eddieantonio:
This allows users to zoom in on content on the mobile site. Previously, the perspective and text sized was fixed, which made for bad user experience when, e.g., a users wants to zoom in on text or a particular image. This PR enables zoom in, while keeping the viewport looking correct on mobile browsers.

Related reading:
 - http://adrianroselli.com/2015/10/dont-disable-zoom.html
 - https://stackoverflow.com/a/22544312/6626414
 - https://sitebulb.com/hints/mobile-friendly/the-viewport-meta-tag-prevents-the-user-from-scaling/

### DIFF
--- a/_includes/includes/head.php
+++ b/_includes/includes/head.php
@@ -26,7 +26,7 @@
 <head>
   <meta charset="utf-8">
   <title><?php echo $title; ?></title>
-  <meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=1.0">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <link rel='shortcut icon' href="<?php echo $favicon; ?>">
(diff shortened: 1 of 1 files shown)


mcdurdin:
Question: do we need to use [`shrink-to-fit=no`](http://adrianroselli.com/2015/10/dont-disable-zoom.html)? Is that something you can verify before merge?

eddieantonio:
We don't _need_ to use it, but on the off-chance we have content that spills out of the right of the viewport, and somebody is viewing the page on iOS 9.0–9.2, this fixes it. It seems the behaviour this declaration addresses is actually a regression introduced in iOS 9, and not a genuine feature!

PR #48, opened 2019-09-20
Add guide for distributing lexical model packages

darcywong00:
This adds a page for distributing lexical model packages. 

It's a little WET with keyboard package content. I thought about combining the guides, but decided the variations in keyboard vs lexical-models would confuse users.

* Add some reference file types for lexical models

Note: a lexical model package creation guide will be handled on a separate PR. (That link would get added to `developer/12.0/guides/lexical-models/distribute/index.php`)


developer/12.0/guides/lexical-models/distribute/packages.php
```diff
@@ -0,0 +1,72 @@
...
+
+  head([
+    'title' => "Distribute Lexical Models to Keyman Applications"
+  ]);
+?>
+
+<h1 class="title" id="reference__package">Distribute lexical models to Keyman Applications</h1>
+<h2>Overview</h2>
+
+<p>Lexical model package files contain one or more lexical models, along with readme files,
```
jahorton: > or more

I thought we only packaged a single model within a lexical model package?  It may target more than one language tag ('en' vs 'en-us' vs 'en-uk'), but it's just for a single model.

darcywong00: In the Developer Package editor, you can add multiple models into a package.
Hmm, dunno what the mobile apps will actually do with that though :)

jahorton: > In the Developer Package editor, you can add multiple models into a package.

@mcdurdin Is that intended?

darcywong00: kmp.json looks valid:
(I didn't add a description to MTNT)

```
  "lexicalModels": [
    {
      "name": "sencoten",
      "id": "nrc.str.sencoten",
      "version": "1.0",
      "languages": [
        {
          "name": "Straits Salish",
          "id": "str"
        }
      ]
    },
    {
      "name": "",
      "id": "nrc.en.mtnt",
      "version": "1.0",
      "languages": [
        {
          "name": "English",
          "id": "en"
        }
      ]
    }
  ]
```

eddieantonio: Just because it's technically valid, doesn't mean it's necessarily a good idea :/ I worry that distributing multiple models per package may cause a lot of confusion, both for model developers and end users.

mcdurdin: It's technically valid but not recommended at this time. The design doesn't restrict it but our guidelines will. The lexical-models repo requires one model per package. I took this approach because it leaves the door open from a design perspective for methods of distribution we haven't yet considered.

developer/12.0/guides/lexical-models/distribute/packages.php
```diff
@@ -0,0 +1,72 @@
...
+
+<dl>
+  <dt>*.model.js</dt>
+  <dd>Lexical model files. When Keyman mobile applications install a lexical model package, all included models will be installed as a group.</dd>
+
+  <dt>welcome.htm</dt>
+  <dd>Introductory help for the lexical model, HTML format. This will normally be displayed when the package is installed by the user, and is also the entry point
+  for help when accessed via Keyman's help system or Keyman Configuration.</dd>
+
+  <dt>kmp.json</dt>
```
mcdurdin: This kmp.json section is a little confusing to include here because it's internal data that's generated at compile time. You can't see it in the source editor. It doesn't need to be in the guide, just in the file format reference for .kmp files.

developer/12.0/guides/lexical-models/distribute/packages.php
```diff
@@ -0,0 +1,72 @@
...
+</dl>
+
+<h2>Step 1) Share the lexical model package file</h2>
+
+<p>Once the lexical model package .kmp file is created, you can share them via external storage devices (USB drive, SD card, etc).
+  If that is not an option, you can upload the .kmp file to a public facing website.
+  For this example, the lexical model package for SENĆOŦEN is being uploaded:</p>
+
+<ol>
+  <li>nrc.str.sencoten.kmp (the lexical model package .kmp file)</li>
```
mcdurdin: Should be called `nrc.str.sencoten.model.kmp` really. We don't enforce that naming convention but it helps to be consistent in our examples.

developer/12.0/reference/file-types/model-js.php
```diff
@@ -0,0 +1,26 @@
...
+<dl>
+  <dt>Used by:</dt>
+  <dd><span class="application">Keyman for Android</span> and <span class="application">Keyman for iPhone and iPad</span>.</dd>
+
+  <dt>Description:</dt>
+  <dd>A .MODEL.JS file is a compiled lexical model file.</dd>
+
+  <dt>Details:</dt>
+  <dd>A .MODEL.JS file is compiled in <span class="application">Keyman Developer</span> from the Keyman lexical model
+    source (<a href='model-ts'>.MODEL.TS</a>) file and (<a href="tsv">.TSV</a>) wordlist. Note that .MODEL.JS files will only work in
```
mcdurdin: Change `will only work` to `are currently only supported`

developer/12.0/reference/file-types/model-js.php
```diff
@@ -0,0 +1,26 @@
...
+  <span class="application">Keyman for Android</span> and <span  class="application">Keyman for iPhone and iPad</span>.</dd>
+
+  <dt>Distributed with lexical model:</dt>
+  <dd>This is the lexical model file. It must be distributed with your package to allow people to use the lexical model
```
mcdurdin: `s/with your package/as a part of your package`

developer/12.0/reference/file-types/model-ts.php
```diff
@@ -0,0 +1,21 @@
...
+?>
+
+<h1>MODEL.TS files</h1>
+
+<dl>
+  <dt>Used by:</dt>
+  <dd><span class="application">Keyman Developer</span>.</dd>
+
+  <dt>Description:</dt>
+  <dd>A .MODEL.TS file is a lexical model definition source file. This holds all the code used by a Keyman lexical model, in plain text.</dd>
```
mcdurdin: Add information about the language of the source file, TypeScript, with a link to http://www.typescriptlang.org/.

developer/12.0/guides/lexical-models/distribute/packages.php
```diff
@@ -0,0 +1,72 @@
...
+  head([
+    'title' => "Distribute Lexical Models to Keyman Applications"
+  ]);
+?>
+
+<h1 class="title" id="reference__package">Distribute lexical models to Keyman Applications</h1>
+<h2>Overview</h2>
+
+<p>Lexical model package files contain one or more lexical models, along with readme files,
+  and any other files you wish to include. You should create a package file to bundle your lexical models and help
```
mcdurdin: `s/should/must/

developer/12.0/guides/lexical-models/distribute/packages.php
```diff
@@ -0,0 +1,72 @@
...
+  into a simple, single file that is easy for an end-user to install.</p>
+
+<p>Internally, as well as your own files, the package file will contain a metadata files "kmp.json", which lists the details Keyman
```
mcdurdin: I'm not sure this information is helpful here? It belongs more in the file format reference for .kmp. In any case `s/a metadata files/a metadata file/`

developer/12.0/guides/lexical-models/distribute/packages.php
darcywong00: got it. I'll update the guide to instruct one model per package

developer/12.0/guides/lexical-models/distribute/packages.php
darcywong00: kmp.json removed

developer/12.0/reference/file-types/model-ts.php
darcywong00: Added to Details

PR #51, opened 2019-09-25
Android main help page:  WET -> DRY with CSS

jahorton:
Before going further with the Android help page (by splitting it into multiple smaller pages), I felt it wise to focus on a simple redo of the existing page using CSS tricks.  Rather than maintain two separate tabs and performing wholesale "display: none" vs "display: block" operations, this uses CSS to toggle which image is displayed on the page in line with the surrounding text.

In all, this allows us to write the documentation on the page itself just the once, though of course there is all the PHP for image source links.

Note that this does temporarily break the ability to easily offline-mirror the page, but I've got a script near ready to handle that side of things (similar to the one introduced in https://github.com/keymanapp/keyman/pull/2102).

I'm still working on getting the Android app to produce the appropriate 'formFactor' query parameter for use with the page, but I can already see the path forward for it, both for online and offline modes.  In the meantime, the page still supports the old tab-button menu for phone vs tablet images, and their use will set `$_SESSION` accordingly.  Not that it's of use with any other pages _yet_, but it will be.

And, uh, please ignore the description of the second commit.  Kinda got it inverted there.  Third commit is mostly whitespace, so a good ol' `?w=1` may help.

### DIFF
--- a/_includes/includes/session-embed.php
+++ b/_includes/includes/session-embed.php
@@ -23,6 +23,4 @@ $embed_mac     = $embed == 'macos';
 $embed_linux   = $embed == 'linux';
 $embed_android = $embed == 'android';
 $embed_ios     = $embed == 'ios';
-
-// Generates a stylesheet to control visibility based on embedding state
 ?>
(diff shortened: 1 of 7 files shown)


mcdurdin:
> a good ol' `?w=1` may help

Note that this now more accessible with the cog at the top of the `Files changed` page :)

PR #54, opened 2019-09-26
Basic keymanweb.com product page

jahorton:
While the 11.0 section has more content, said content is a duplicate of the also-deprecated web/_`<`version`>`_/ui content - there's no reason to maintain it twice.  This leaves the Products > keymanweb.com section with a pretty small amount of content, but that's probably fine.

### DIFF
--- a/products/web/12.0/index.php
+++ b/products/web/12.0/index.php
@@ -20,7 +20,7 @@
   Bookmarklet for your browser.
 </p>
 
-<p><a href='http://keymanweb.com'>Visit Keymanweb.com now</a></p>
-<a href='http://keymanweb.com'><img class="center" src="<?php echo cdn('img/kmw_screenshot.png'); ?>" /></a>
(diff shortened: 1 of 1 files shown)


products/web/12.0/index.php
```diff
@@ -0,0 +1,26 @@
...
+<p>
+  You can use Keymanweb.com to type in any of our available languages, and then post your message to Twitter,
+  as well as search Google or copy it to your clip board.
+  <br/><br/>
+  Keymanweb.com also provides links to download the active keyboard with Keyman Desktop, or install it as a
+  Bookmarklet for your browser.
+</p>
+
+<p><a href='http://keymanweb.com'>Visit Keymanweb.com now</a></p>
+<a href='http://keymanweb.com'><img class="center" src="<?php echo cdn('img/kmw_screenshot.png'); ?>" /></a>
```
darcywong00: We can use `https://`

PR #95, opened 2019-12-13
fix: deprecated keyboard search

mcdurdin:
Documentation relating to keymanapp/api.keyman.com#28 and keymanapp/keyman.com#62.

### DIFF
--- a/developer/cloud/keyboard_info/1.0/index.php
+++ b/developer/cloud/keyboard_info/1.0/index.php
@@ -336,7 +336,7 @@
 <td>deprecated</td>
 <td></td>
 <td></td>
-<td></td>
+<td>yes</td>
 <td>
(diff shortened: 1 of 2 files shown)


developer/cloud/keyboard_info/1.0/index.php
```diff
@@ -332,6 +332,17 @@
   <p><b>object</b>, with each property an id of a keyboard and corresponding value is a <b>KeyboardRelatedInfo</b>.</p>
 </td>
 </tr>
+<tr>
+<td>deprecated</td>
+<td></td>
+<td></td>
+<td></td>
```
darcywong00:
```suggestion
<td>yes</td>
```

Since the deprecated field is "always generated automatically"

PR #98, opened 2020-01-06
Clean up support details for Developer language

### LINKED ISSUE #97: Add Linux support details to support grids for Developer language reference
The platform support grids in the Developer language reference do not include Keyman for Linux. This should be added.

While you are at it...

## Platform support anomalies
* mnemonic layout support is currently incorrect for Mac. While mnemonic keyboards work on Mac, you cannot select a different base keyboard.

* If statement support is also incomplete as the options form is not supported on mac (or Linux).

darcywong00:
Fixes #97

* Add Linux to the Developer support grids
* Clarify mnemonimc keyboards work on Mac, but cannot select a different base keyboard
* Mark `if()` as incomplete on Mac and Linux because options form is not supported

### DIFF
--- a/developer/language/reference/if.php
+++ b/developer/language/reference/if.php
@@ -60,7 +60,7 @@ if(opt2 = '1') + any(letterGroup1) > index(outputGroup1,2)
 
 <p>The <code>if()</code> statement can be used in keyboards on the following platforms.</p>
 
-<p>Support for Mac and Linux is incomplete as the options form is not supported.</p>
+<p>Support for Mac and Linux is incomplete as the options.htm form to set variable stores is not supported.</p>
 
 <table class='platform'>
   <thead>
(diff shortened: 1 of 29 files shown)


developer/language/reference/if.php
```diff
@@ -60,12 +60,14 @@
 
 <p>The <code>if()</code> statement can be used in keyboards on the following platforms.</p>
 
+<p>Support for Mac and Linux is incomplete as the options form is not supported.</p>
```
jahorton: This part could probably use some clarification.  Looking at the page overall, I wouldn't be sure what parts are and are not the "options form" this part is mentioning.

developer/language/reference/mnemoniclayout.php
```diff
@@ -69,12 +69,14 @@
 key codes in the touch definition. Web keyboards have limitations due to the inconsistent ability to determine the active physical
 layout.</p>
 
+<p>While mnemonic keyboards work on Mac, you cannot select a different base keyboard.</p>
```
mcdurdin: I believe the situation is the same for Linux?

darcywong00: Gotcha. I'll include Linux

PR #99, opened 2020-01-22
feat: edit page link

mcdurdin:
Provides a quick link to the GitHub editor
for the page in question. Hope that this
streamlines quick edits to help pages and
encourages minor fixes by us and others.

### DIFF
--- a/_includes/includes/footer.php
+++ b/_includes/includes/footer.php
@@ -45,8 +45,11 @@
         </div>
     </div>
 </div>
-<div id="jira-feedback">
-  <div id="jira-feedback-tab"><h4><a href='https://community.software.sil.org/c/keyman'>Support</a></h4></div>
+<div class="footer-tab-holder" id='footer-tab-support'>
+  <div class="footer-tab"><h4><a href='https://community.software.sil.org/c/keyman'>Support</a></h4></div>
+</div>
+<div class="footer-tab-holder" id='footer-tab-edit'>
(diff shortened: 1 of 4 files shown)


_includes/includes/footer.php
```diff
@@ -45,8 +45,11 @@
         </div>
     </div>
 </div>
-<div id="jira-feedback">
-  <div id="jira-feedback-tab"><h4><a href='https://community.software.sil.org/c/keyman'>Support</a></h4></div>
+<div class="footer-tab-holder" id='footer-tab-support'>
+  <div class="footer-tab"><h4><a href='https://community.software.sil.org/c/keyman'>Support</a></h4></div>
+</div>
+<div class="footer-tab-holder" id='footer-tab-edit'>
+  <div class="footer-tab"><h4><a href='https://github.com/keymanapp/help.keyman.com/edit/master<?php echo $_SERVER['PHP_SELF'] ?>' target='_blank'>Edit page</a></h4></div>
```
darcywong00: This works out nicely!

Do you want to exclude some of the auto-generated pages? (e.g. keyboards or Desktop)

mcdurdin: Yeah, I wondered about that ... it makes sense from our workflow not to edit the generated files, but if an end user makes a contribution, we can always port their PR to the correct location. So, for now let's leave it on and see what happens: if this becomes popular, we can add appropriate workflows for those changes too (i.e. for desktop, load the appropriate file in keyman repo, and for keyboards, load the appropriate file there).

Note: for us, we get the option in the editor to try to merge to master (which will fail obvs) or create a PR; other users will only be able to create a PR.

PR #112, opened 2020-02-14
chore: cleanup of css

mcdurdin:
Adjusts the styling of text, the layout and widths of columns, index, and toc, and makes better use of screen real estate. Lots of subtle things.

Also, fixes the bouncing TOC issue where the "On this page" section on the right would jump left or right as you scrolled.

Removing index.css which was now completely redundant, after verification.

From there, I adjusted a bunch of CSS around the product indexes and fixed up margins and little bits and pieces like that.

Also removed redundancy in font specifications. The fonts will need further work one day.


cdn/dev/css/fonts.css
```diff
@@ -0,0 +1,143 @@
...
+/* Marathi
+:lang(mar).lang-example {
+	font-family: GeezWeb;
+} */
+/* Oriya */
+:lang(ori).lang-example {
+	font-family: OriyaWeb;
+}
+/* Rawang */
+:lang(raw).lang-example {
```
darcywong00: Note for keymanapp/keyman.com#65 I had to comment out Rawang for it to render correctly on the footer

darcywong00:
Viewing from a mobile viewport, the homepage styling seems off

WIn 10, Chrome
help-mobile-view

darcywong00:
Should the bullets in unordered lists be removed?

http://help.keyman.com.local/developer/engine/android/12.0/
image

mcdurdin:
> Viewing from a mobile viewport, the homepage styling seems off
> 
> WIn 10, Chrome

The watermarking was deliberate ... an attempt to try and keep the image without losing real estate. If you think it doesn't work, I can change it.

> Should the bullets in unordered lists be removed?
> 
> http://help.keyman.com.local/developer/engine/android/12.0/

There is a problem with the automatic index generation in the engine folder -- if I fix that, it'll fix the bullets. The bullets make the index much easier to read (normally, anyway!)

darcywong00:
> The watermarking was deliberate ... 

Ah, yeah that's fine with me then

mcdurdin:
I've tweaked the bullets situation; tidied a few pages. Thought about removing the `$pagetitle` construct altogether but found 1446 results in 1342 files and thought perhaps another day.

PR #116, opened 2020-02-16
feat: markdown support and content versioning

mcdurdin:
This is probably most easily reviewed commit-by-commit. Two significant features:

* Markdown support (b70706c4, dafb2f0e, e617bc85)
* Content versioning (025d8c16, c0eb1f2c)

Significant maintenance:
* refactoring of .php redirection to name-without-extension, from 11 (I think) locations to 1, in root web.config (08f0b5d7),
* a lot of broken links resolved (ba202a13)

# Markdown support

Uses existing parsedown parser. If a file.md exists, it will be used in preference to file.php; this is controlled by root web.config. I've included a couple of examples in commit dafb2f0e.

# Content versioning

This simplifies the maintenance and navigation of multiple versions of content by providing automatic linking between different versions of the same content. An initial implementation is in commit 025d8c16 but the completed implementation is in c0eb1f2c.

## Details

Adds a version dropdown at the top of pages which have multiple versions such as /developer/12.0/ etc.

Where possible, content should refer to current-version instead of a specific version, when referencing cross-scope, e.g. use /developer/current-version/guides/..., not /developer/12.0/guides/....

Root web.config rewrites /current-version/ to /12.0/. When we release, we just update that rewrite to /13.0/. If content is missing for 13.0, it will fall back to earlier versions, saving us from duplicating identical content, e.g. when engine hasn't changed.

Intelligently provides links to the same content in different versions and deals with missing content by linking to home page for the version instead.

Removes hand-coded version links where I found them in preference for the machine generated links.

Versioning applies to /developer and /products paths only at present; the versioning rules are fairly consistent there but much more complicated for /keyboards (and may not be very achievable with the web.config strategy I am applying).

Logic is in /_includes/includes/page-version.php and /web.config.


_includes/includes/index-content.php
```diff
@@ -1,26 +1,32 @@
 <?php
 
-function get_file_titles($path) {
-  $files = glob($path.'*');
+function get_file_titles($path,$base) {
+  $files = glob($base.'*');
```
jahorton: Why are we keeping the `$path` variable around?  It seems like it's no longer utilized at all, which makes me think something (like defaulting `$base=$path` if `$base` is empty) is missing.

_includes/md/mdhost.php
```diff
@@ -0,0 +1,41 @@
...
+    die('Require file parameter');
+  }
+
+  $file = realpath(__DIR__ . '/../../') . DIRECTORY_SEPARATOR . $_REQUEST['file'];
+  $contents = trim(file_get_contents($file));
+  $contents = str_replace("\r\n", "\n", $contents);
+
+  if(preg_match('/^---\n(.+)\n---\n(.+)/s', $contents, $match)) {
+    $metadata = $match[1];
+    $contents = $match[2];
```
jahorton: Do we have any documentation somewhere regarding this format?  I assume this is meant to interact with the `title:` check in _includes/includes/index-content.php.

If I'm parsing this right, a section like the following will be auto-omitted from `$contents`, which is the part of the file that's passed to Parsedown?

```
---
title: Hello World
platform: web
---
```

web.config
```diff
@@ -12,23 +12,175 @@
...
+                  </conditions>
+                  <action type="Redirect" url="{R:1}" />
+                </rule>
+
+                <rule name="Rewrite file to file.md" stopProcessing="true">
+                  <match url="^(.+)$" />
+                  <conditions logicalGrouping="MatchAll" trackAllCaptures="false">
+                    <add matchType="IsFile" input="{DOCUMENT_ROOT}/{R:1}.md" />
+                  </conditions>
+                  <action type="Rewrite" url="/_includes/md/mdhost.php?file={R:1}.md" />
```
jahorton: So this is what performs the .md rendering magic, right?

_includes/md/mdhost.php
mcdurdin: Added some documentation in the next commit.

web.config
mcdurdin: Correct.

mcdurdin:
> Shouldn't we instead use descending order? The more current the available version, the more likely the user wants help regarding that version, users typically read top-to-bottom, and it makes the mouse-click position more consistent across the platforms, too. (The 'desktop' dropdown, in particular, will take quite a lot of reading to find 12.0 or 13.0.)

Agreed. Resolved.

> Secondary concern: I see that Desktop has an entry for 14.0 already. Are we fine with that being present, or might we want the option to keep certain versions off the list, like for extreme early alphas?

I am actually really happy if people are curious about future functionality. It's not like we want it to be a secret, so while there is some minor potential for confusion, for most users I don't think it will be a concern 😊

> ... and for the rest, I take it the `/\/old\//` and `/(.*Deprecated)/` file deletions are confirmed no-longer accessible documentation files that have been fully replaced since they were marked as such?

Yes, most of those were only accessible due to automated index references to the files, and they contained invalid content anyway! So I removed them.

> That commit ([ba202a1](https://github.com/keymanapp/help.keyman.com/commit/ba202a1316fa83d29994aaeec0186f988ceda8b1)) is a bit much to sift through personally with a fine-tooth comb, so if you say you've investigated it and ensured things are handled properly there, I'm fine rolling with that.

I used a link testing tool to verify the links on the site (about 18,000 internal URLs, plus 5833 external URLs linked to which were not checked, took 5 minutes on my local clone of the site). There were hundreds of internal broken links, most of which I have resolved now. There are about 100 internal broken links left, most in very old documentation. I'd like to clean those up at some point too, but out of time for now.

developer/engine/desktop/web.config
```diff
@@ -4,45 +4,17 @@
         <rewrite>
             <rules>
                 <!-- Desktop engine content hasn't changed between 10.0 -> 12.0 -->
-                <rule name="Redirect 11.0/ and 12.0/ to 10.0/">
+                <rule name="Redirect 11.0/ and 12.0/ to current-version/">
                     <match url="(11|12)\.0(.*)" />
```
darcywong00:
```suggestion
                    <match url="(1\d).0(.*)" />
```
so /13.0/ (and above) will redirect to current_version

developer/engine/desktop/web.config
```diff
@@ -4,45 +4,17 @@
...
-                    <action type="Redirect" url="10.0{R:2}" />
+                    <action type="Redirect" url="current-version{R:2}" />
                 </rule>
 
-                <rule name="Redirect unversioned paths to 10.0/ path">
+                <rule name="Redirect unversioned paths to current-version/ path">
                     <match url="^(.*)" />
                     <conditions>
-                        <add input="{URL}" pattern="(9|10)\.0(\/)?" negate="true" />
+                        <add input="{URL}" pattern="(current-version|\d+\.\d+)(\/)?" negate="true" />
```
darcywong00: Where to redirect http://help.keyman.com.local/developer/engine/desktop/8.0/ ?

developer/engine/iphone-and-ipad/13.0/Manager/showKeyboardPicker.php
```diff
@@ -42,5 +42,5 @@
 
 <h2 id="See_also" name="See_also">See also</h2>
 <ul>
- <li><a href="showLanguageList.php"><code>showLanguageList()</code></a></li>
+ <li><a href="showKeyboardPicker.php"><code>showKeyboardPicker()</code></a></li>
```
darcywong00: Can remove self-referencing link (for iphone-and-ipad/10.0 to 13.0)

developer/engine/desktop/web.config
mcdurdin: Hmm. Actually, I wonder if we should no longer have the 11.0 and 12.0 redirects, because we aren't forcing a version update for each release now with the new pattern? But happy to go with your suggestion for now.

developer/engine/desktop/web.config
mcdurdin: That one should 404.

darcywong00:
With this PR, should future keymanapp/keyboards PRs use markdown for documentation instead of help php?

mcdurdin:
> With this PR, should future keymanapp/keyboards PRs use markdown for documentation instead of help php?

Not yet. I don't think it reduces the workload for keyboard developers given welcome.htm is still in HTML. Something to consider for the future for sure.

developer/cloud/1.0/index.md
```diff
@@ -0,0 +1,5 @@
+---
+title: KeymanWeb Server Data API 1.0
```
darcywong00: When I tried the blue "Edit page" link, I got a 404 for
https://github.com/keymanapp/help.keyman.com/edit/master/developer/cloud/1.0/index.md

mcdurdin: Well yes, because that page is not yet merged to master 😁

PR #120, opened 2020-02-16
Change font to distinguish kbd styling of i, I, l, 1

### LINKED ISSUE #96: kbd styling doesn't differentiate between l (small L) and I (capital i)
As seen at https://help.keyman.com/keyboard/itrans_odia/1.0.0/itrans_odia.php, the CSS styling for kbd elements uses 'sans-serif' as the font, which doesn't distinguish between `l` and `I`:

image

darcywong00:
Fixes #96 

(updated for NotoSans-Regular-Ascii)
For `<kbd>` styling, replace current sans-serif with NotoSans-Regular-Ascii so `i`, `I`, `l`, `1` are distinguishable. Due to the original size of NotoSans-Regular, the Ascii subset is used for this PR.

### NotoSans (Ascii portion)
notosansregular-ascii

### Calibri, sans-serif (for comparison)
calibri

http://help.keyman.com.local/developer/13.0/guides/develop/tutorial/step-7 
uses `<kbd>I</kbd>`

Note: the styling mentioned in itrans_odia from the issue is maintained in the keyboards repo, 
Per @mcdurdin 
> We won't worry about those for now


mcdurdin:
NotoSans-Regular is very large (445kb). Is there a smaller version that we could use instead? We only need U+0033-U+007E!

darcywong00:
> Is there a smaller version that we could use instead? We only need U+0033-U+007E!

Going by the character counts on 
https://en.wikipedia.org/wiki/Noto_fonts#List_of_Noto_fonts

I tried the smallest (Noto Sans Tagbanwa), but the "1" isn't as pronounced
image

darcywong00:
Python script to extract Ascii characters from NotoSans-Regular
```
pyftsubset ../fonts/Noto_Sans/NotoSans-Regular.ttf --unicodes=U+0020-007F
```

PR #149, opened 2020-03-06
fix: support keyboard fonts

### LINKED ISSUE #150: Keyboard renderer does not currently load webfonts as specified in .keyboard_info
The keyboard renderer should use the information provided in the .keyboard_info to present osk.

For example, see https://help.keyman.com/keyboard/newa_traditional/1.0/newa_traditional. The web font is available on s.keyman.com.

This issue was blocked by the keyboard using a .otf. Keyman 13.0.101 adds support for .otf extension but the help site is currently using 10.0.103 (!) and updating to 13.0.101 breaks the site in other ways.

This may be accomplished by using the KeymanWeb API's in a more appropriate manner or by consuming the data directly from api.keyman.com -- will defer to @jahorton.

Depends on keymanapp/keyman#2784, keymanapp/keyman#2818.

mcdurdin:
This is a quick fix to get keyboard fonts working
on help pages; it relies on api.keyman.com,
s.keyman.com being in sync. In future we should
look at leveraging KeymanWeb's built-in font
support but that requires more re-engineering
of the help kbd-docs js.

Fixes #150.

### DIFF
--- a/cdn/dev/js/kbd-docs.js
+++ b/cdn/dev/js/kbd-docs.js
@@ -10,11 +10,11 @@ function loaded(){
   $('#testimonial').click(function(){
     location.href="http://www.tavultesoft.com/testimonials.php";
   });
-  
+
   // Email subscribe form
   $('.subscribe').click(function(){
     $('#mc-embedded-subscribe-form').submit();
-  });    
+  });
 
   // Popup close
   $('#popup-close').click(function(){
@@ -32,9 +32,9 @@ function loaded(){
 	$('.popup').fadeOut(300);
     }
   });
-  
+
   // Scrolling top menu functionality (Non touch devices only)
-  var 
+  var
     device = $(document.body).data('device'),
     scrollFix = function(e) {
       var scroller_anchor = $(".header").height();
@@ -48,7 +48,7 @@ function loaded(){
         $('#toc').removeClass('fixed');
       }
     };
-    
+
   if (device != 'Android' && device != 'iPhone' && device != 'iPad') {
     $(window).scroll(scrollFix);
   }
@@ -61,10 +61,10 @@ function loaded(){
       }else{
         $("body").css("overflow","auto");
         $("#phone-menu").hide();
-      }	
+      }
     });
   }
-  
+
   // Footer to bottom of page
   var link = $('.footer');
   var offset = link.offset().top;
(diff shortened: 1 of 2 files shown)


cdn/dev/js/kbd-docs.js
```diff
@@ -61,30 +61,61 @@ function loaded(){
...
+    // This retrieves font info from latest version of keyboard; we don't currently have that available for earlier versions. No worries.
+    $.ajax({
+      url: 'https://api.keyman.com/keyboard/' + keyboardName,
+      success:function(data) {
+        if(typeof data != 'object') return;
+        for(var lang in data.languages) {
+          if(data.languages[lang].font) {
+            // pick the first font
+            fontFamily = data.languages[lang].font.family;
+            fontSource = data.languages[lang].font.source;
```
darcywong00: I get a 404 on http://help.keyman.com.local/keyboard/burmese02/1.1/burmese02

Looks like the source can be an array
```
    "languages": {
        "my": {
            "font": {
                "family": "MyanmarWeb",
                "source": [
                    "Padauk.ttf",
                    "PADAUK0.eot",
                    "Padauk.mobileconfig"
                ]
            },
```
and the .ttf might not always be listed first?

mcdurdin: Good point. 

It’s possible this PR isn’t needed. It may be resolved with the update to KMW. Will wait and see.

darcywong00: Well, the PR definitely works for http://help.keyman.com.local/keyboard/newa_traditional/1.0/newa_traditional

The OSK seems to get loaded twice though?

mcdurdin: Okay, resolved I think. Not sure what you mean by: 

> The OSK seems to get loaded twice though?

darcywong00: http://help.keyman.com.local/keyboard/burmese02/1.1/burmese02 seems to render the keyboard twice

Initial (currently matches live site)
unshifted 1

Final (slightly thinner font)
unshifted 2

mcdurdin: Okay, I think that's not an issue -- the initial load uses the font on your machine, then when the webfont loads, it applies it.

mcdurdin:
Not merging until keymanapp/keyman#2807 is resolved: it breaks keyboard display entirely at present.

PR #179, opened 2020-05-05
Update common_requirements.php

DavidLRowe:
This has been "Keyman Desktop 10.0" for a number of versions. Why not drop the version number?


mcdurdin:
Unfortunately, you've managed to hit another of our edit page gaps -- the documentation for Keyman Desktop is still sourced from the Keyman repository rather than the other way around. All other projects take a copy of the documentation from help.keyman.com at build time, but Keyman Desktop's docs are at https://github.com/keymanapp/keyman/tree/master/windows/src/desktop/help :(

PR #190, opened 2020-07-08
Update initialize.php

mcdurdin:


developer/engine/android/13.0/KMManager/initialize.php
```diff
@@ -35,7 +35,6 @@
 protected void onCreate(Bundle savedInstanceState) {
     super.onCreate(savedInstanceState);
     // ...
-    KMManager.setKeymanLicence("********", "********");
```
darcywong00, line 38: This is also in 10.0, 11.0 (interestingly not 12.0)

darcywong00: ack! I meant to handle 10.0, 11.0, and 14.0 (ac6644dbfc0cd89007d2b258c03d645911c500b4) on a separate branch and make a PR. But accidentally pushed to master. :/  (dunno how that worked...)

mcdurdin: As you are an admin on the repo, you have permission to push to master.

PR #193, opened 2020-07-09
Add DefaultKeyboard() and deprecate showLanguageList

darcywong00:
Document the KMEA KMManager API changes from keymanapp/keyman#3323

Also mark showLanguageList() as deprecated


darcywong00:
Noted. I've updated the description on `showLanguageList()`, and also added more `(Deprecated)` markings to other calls.

PR #199, opened 2020-07-22
Update stores.php

sewhite:
Rewording this to be more clear (following the discussion in the community site).
"a 1-based offset" is totally obscure to the average user.

### DIFF
--- a/developer/language/guide/stores.php
+++ b/developer/language/guide/stores.php
@@ -51,8 +51,10 @@ context or key section of the rule, and will match any character in the specifie
 </code></pre>
 
 <p>The <code>index()</code> statement is used to look up a character in a store at the index where a specified <code>any()</code> statement matched
-a character. The <code>any()</code> statement used to obtain the index is referenced by its position in the context (and key) part of the rule,
-a 1-based offset.</p>
+a character. The <code>any()</code> statement used to obtain the index is referenced by its position in the context (and key) part of the rule. 
+In the first example line,The <code>any()</code> statement is the first element (character or statement) in the rule, so the offset is 1. In  the second example line, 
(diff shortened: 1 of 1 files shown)


developer/language/guide/stores.php
```diff
@@ -51,8 +51,10 @@
 </code></pre>
 
 <p>The <code>index()</code> statement is used to look up a character in a store at the index where a specified <code>any()</code> statement matched
-a character. The <code>any()</code> statement used to obtain the index is referenced by its position in the context (and key) part of the rule,
-a 1-based offset.</p>
+a character. The <code>any()</code> statement used to obtain the index is referenced by its position in the context (and key) part of the rule. 
+In the first example line,The <code>any()</code> statement is the first element (character or statement) in the rule, so the offset is 1. In  the second example line, 
+if the three dots represent one single character or statement, the <code>any()</code> is the second element of the rule so the offset is 2.  Since the context part of a rule
+can contain several characters or statements, the offset for the <code>any()</code> could be greater than two.
```
mcdurdin, line 57:
```suggestion
a character. The <code>any()</code> statement used to obtain the index is referenced by its character position in the context (and key) part of the rule. 
In the first example line, the <code>any()</code> statement is the first element (character or statement) in the rule, so the offset is 1. In 
the second example line, if the three dots represent one single character or statement, the <code>any()</code> is the second element of the
rule so the offset is 2.  Since the context part of a rule can contain several characters or statements, the offset for the <code>any()</code>
could be greater than two. 
```

I'm not sold on the description around the 'three dots' -- I think it may end up confusing things further, given that is pseudocode. I think we might be better off with more examples rather than attempting to explain that :)

PR #202, opened 2020-07-23
feat: add tests and traffic cone

mcdurdin:
Traffic cone appears on staging and development sites. Just a little hint that we are not browsing the production server.

The broken link test currently returns 1294 broken internal links out of 1,157,897 detected links(!) and I think we can fix these as we go. This means that many PRs will return failed status checks but I think we probably just ignore that for now. One lazy day soon I'll grab the log and knock off the low-hanging-fruit broken links.


darcywong00:
Looks like the lint action failed on the PR build

mcdurdin:
Lint now passes, so it's just broken links which I will be resolving later.

keyboard/clavbur9/9.1/clavbur9.php
```diff
@@ -1,5 +1,5 @@
 <?php
-  $pagename = 'Guide d'utilisation pour le clavier du Burkina / Clavier du Burkina Keyboard Help';
+  $pagename = "Guide d'utilisation pour le clavier du Burkina / Clavier du Burkina Keyboard Help";
```
darcywong00, line 2: Won't the change from 4bc979027b554207edb76a20981623300505d044 need to be done in the keyboards repo?

mcdurdin: See https://github.com/keymanapp/keyboards/pull/1296. This mitigates for staging so the tests pass (I may need to resolve conflict later I guess, no big deal).

PR #217, opened 2020-09-16
fix: Update kmx usage

### LINKED ISSUE #197: Update `kmx.php`
`kmx.php` mentions that `.kmx` files are only used and will only work in Keyman Desktop which is no longer true.

darcywong00:
Fixes #197

Starting with 11.0, Keyman for Linux also uses kmx.


mcdurdin (approved):
LGTM. FWIW, we don't normally update older versions of documentation, only versions under maintenance.

PR #229, opened 2020-09-25
fix(developer): broken links

darcywong00:
Fix some more developer links
Seed core/desktop/14.0 content


developer/engine/android/13.0/KMManager/getKeyboardFontFilename-Deprecated.php
```diff
@@ -29,6 +29,6 @@
 
 <h2 id="See_also" name="See_also">See also</h2>
 <ul>
- <li><a href="getKeyboardFontTypeface.php"><code>getKeyboardFontTypeface()</code></a></li>
+ <li><a href="getKeyboardFontTypeface-Deprecated.php"><code>getKeyboardFontTypeface() (Deprecated)</code></a></li>
```
mcdurdin, line 32: Can we avoid using `-Deprecated` in the filename? Also, shouldn't need to include `.php` in the URL

mcdurdin (changes requested):
I'd really like to avoid using `-Deprecated` in the filenames of the files; also don't include `.php` extension in `href`s. 😉

mcdurdin (approved):

Issue #233, opened 2020-09-29
Remaining broken links to address


mcdurdin:
These were collected in #232 CI run:

- [ ] [developer/cloud/1.0/webapi](https://help.keyman-staging.com/developer/cloud/1.0/webapi)
    * [developer/cloud/current-version/webapi](https://help.keyman-staging.com/developer/cloud/current-version/webapi) (HTTP_404)

---

- [ ] [products/desktop/current-version/docs/context_langsetup](https://help.keyman-staging.com/products/desktop/current-version/docs/context_langsetup)
    * [products/desktop/current-version/docs/advanced_language_config.php](https://help.keyman-staging.com/products/desktop/current-version/docs/advanced_language_config.php) (HTTP_404)
    * [products/desktop/current-version/docs/advanced_language_config.php](https://help.keyman-staging.com/products/desktop/current-version/docs/advanced_language_config.php) (HTTP_404)

---

- [ ] [keyboard/gff_tir_et_7](https://help.keyman-staging.com/keyboard/gff_tir_et_7)
    * [keyboard/gff_tir_et_7/1.0/TigrignaEtTyping-English.pdf](https://help.keyman-staging.com/keyboard/gff_tir_et_7/1.0/TigrignaEtTyping-English.pdf) (HTTP_404)

---

- [ ] [keyboard/gff_tir_er_7](https://help.keyman-staging.com/keyboard/gff_tir_er_7)
    * [keyboard/gff_tir_er_7/1.0/TigrignaErTyping-English.pdf](https://help.keyman-staging.com/keyboard/gff_tir_er_7/1.0/TigrignaErTyping-English.pdf) (HTTP_404)

---

- [ ] [products/desktop/12.0/docs/context_langsetup](https://help.keyman-staging.com/products/desktop/12.0/docs/context_langsetup)
    * [products/desktop/12.0/docs/advanced_language_config.php](https://help.keyman-staging.com/products/desktop/12.0/docs/advanced_language_config.php) (HTTP_404)
    * [products/desktop/12.0/docs/advanced_language_config.php](https://help.keyman-staging.com/products/desktop/12.0/docs/advanced_language_config.php) (HTTP_404)

---

- [ ] [products/desktop/11.0/docs/context_langsetup](https://help.keyman-staging.com/products/desktop/11.0/docs/context_langsetup)
[...]

PR #234, opened 2020-09-29
fix: current-version breaking some links

mcdurdin:
Relates to #233.

The oldest version was selected instead of the newest version in one place, where other places expected the newest version, leading to broken links appearing on some pages, e.g. developer/cloud/1.0/webapi (note, this was appearing in the `<link rel=canonical>` tag, so not visible on the page).


darcywong00 (approved):
lgtm

PR #228, opened 2020-09-25
fix: Update KeyboardEventHandler

darcywong00:
Split out the two interfaces, and add `onPackageInstalled` and `onLexicalModelInstalled`

PR #242, opened 2020-10-07
keyboard submission via GitHub

DavidLRowe:
This is a draft of submitting a keyboard via GitHub.

Comments, corrections, revisions welcome.


developer/keyboards/index.php
```diff
@@ -81,15 +81,21 @@
...
-
+    <li><a href="github">GitHub Keyboard Submission Guide</a></li>
     <p>
-        For advanced users who are managing a larger number of keyboards, learn how to set up and contribute with
-        the full keyboards repository.
+        For users who are creating and submitting a single keyboard or would rather use the GitHub interface,
+        learn how to get set up and contribute with using the online GitHub interface.
     </p>
-    <li><a href="advanced">Advanced Keyboard Submission Guide</a></li>
+    <li><a href="starter">Starter Guide (deprecated)</a></li>
```
darcywong00, line 94: On the `developer/keyboards/starter/index.php` page,
* maybe also add "(deprecated)" 
* and add a link to the github guide

darcywong00:
Thanks for making this guide! Hopefully, it will lessen your pain on keyboard submissions 😄 

btw, I think @mcdurdin is preferring new content to be written in Markdown instead of php.
Here's a link on how to format the titles
https://github.com/keymanapp/help.keyman.com/pull/228#issuecomment-699765788

mcdurdin:
Thank you David!

> btw, I think @mcdurdin is preferring new content to be written in Markdown instead of php.

Yes, I am annoying like this. Markdown is a much nicer format than HTML for this kind of document and makes maintenance much easier in the future. There are tools for converting automatically, such as https://www.browserling.com/tools/html-to-markdown -- just paste the HTML section, not the PHP section.

The thing that you will need to do yourself is insert the right metadata at the top, as described in the link Darcy gave, viz: there's a standard .md metadata section you can add:

```
---
title: my super web page
---
```

For an example, see https://raw.githubusercontent.com/keymanapp/help.keyman.com/staging/developer/cloud/package-version/1.0/index.md

DavidLRowe:
Yes, you had mentioned using markdown, but by the time I got to this task I had forgotten and just copied files from the advanced folder.

I'll update the new files accordingly.

What about existing files? Are you planning to change them to markdown as well?

How do I link from a php file to a md file? (And from a md file to another md file?) Do I need to replace index.php with index.md and work down?

mcdurdin:
> What about existing files? Are you planning to change them to markdown as well?

We'll change files over time as we edit them -- not planning to do a major shift.

> How do I link from a php file to a md file? (And from a md file to another md file?) Do I need to replace index.php with index.md and work down?

* To link to other pages, just don't include the file extension; e.g. if the target is called /developer/foo.php, then use `<a href='/developer/foo'>`.

* For index.php or index.md, leave off the whole filename, for example if you have /developer/creating-emoji-keyboards/index.md, you would reference that as `<a href='/developer/creating-emoji-keyboards/'>`.

* Finally, also use relative paths as far as possible in links (e.g. use `foo` instead of `/developer/foo` if you can, or `../developer/foo` to go up a level). For a relative reference to index use `<a href="index">` (which will redirect silently to the folder URL)

DavidLRowe:
What about PHP include statements, for example: 
```
<?php
    include('../_shared/development-tools.php');
?>
```

mcdurdin:
> What about PHP include statements, for example:

No, you can't use PHP include statements. The content of .md files is intended to be static without server-side changes, which makes them more stable long term and easier to maintain.

DavidLRowe:
So, copy the information from the include into the file?

mcdurdin:
Yes, the file needs to be static. 

Duplicate content is less of a problem here than having more complicated backend configuration. This is generally the case for documentation -- if we have a lot of duplicate content, that's an argument for restructuring the content, not using back-end techniques to copy the same content to multiple pages.

DavidLRowe:
This ended up being trickier than it looked. I also corrected some errors in the original .php files and did some shuffling of content.

Needs more than a cursory review, including testing all the links!

mcdurdin:
> This ended up being trickier than it looked. I also corrected some errors in the original .php files and did some shuffling of content.

Thank you! It's much appreciated. We automatically check links when we merge to the staging server so will do so then, if that's okay.

darcywong00:
On the /developer/keyboards/advanced/index.php page, there's still some mentions/links to the "Starter guide" that you can change to the GitHub guide. (editing the php should be fine, no need to convert that to .md)

developer/keyboards/github/index.md
```diff
@@ -0,0 +1,26 @@
...
+This guide is intended for keyboard developers who are submitting only one or two keyboards. 
+It uses the online GitHub interface. 
+If using "git" from the command line is not something familiar, this is the guide for you. 
+
+Those familiar using "git" from the command line might want to consider the [Advanced Guide for Submitting Keyboards](../advanced). 
+The Advanced guide provides separate instructions for deeper involvement in submitting, improving, and reviewing keyboards.
+
+**Note:** This guide is written for Windows users, so other OS users need to adjust paths accordingly.
+
+<dl>
```
darcywong00, line 16: I think the `<dl>` and `<dt>` tags are interfering with the links. You can remove them (keep the empty lines between the steps).

developer/keyboards/github/step-0.md
```diff
@@ -0,0 +1,154 @@
+---
+title: Preliminary Step: Is Your Keyboard Ready?
+---
+
+# Preliminary Step: Is Your Keyboard Ready?
```
darcywong00, line 5: The `title:` formatting already generates this title on the page, so this line can be removed.
Also applies to each  step-*.md in the guide.

developer/keyboards/github/step-0.md
```diff
@@ -0,0 +1,154 @@
...
+
+<dt>`extras\`</dt>
+
+<dd>other related files, e.g. Word documents, databases, experimental files</dd>
+
+<dt>`source\`</dt>
+
+<dd>keyboard source files, including any files that will be packaged in the build</dd>
+
+<dt>`source\help\<keyboard>.php`</dt>
```
darcywong00, line 56: We may need to investigate how markdown with `<dl><dt> <dd>` tags. In my browser, the `<keyboard>` is treated as an invalid tag so the page renders
> source\help\\.php

mcdurdin: HTML snippets within Markdown cannot themselves contain Markdown. Best to avoid HTML altogether if we can.

Fortunately, with a minor update to our Markdown parser, we can support definition lists (⬅ this makes me Happy Marc).

To do a markdown definition list:

```
`build\`
  : not stored in the repository; output files will go here

`extras\`
  : other related files, e.g. Word documents, databases, experimental files
```

(The indent is not necessary before the colon, but it may make it easier to read the source).

developer/keyboards/github/index.md
```diff
@@ -0,0 +1,26 @@
...
+
+<dt>[Preliminary Step: Is Your Keyboard Ready?](step-0)</dt>
+
+<dt>[Step 1: Setup - One Time Only](step-1)</dt>
+
+<dt>[Step 2: Organizing the Keyboard Files](step-2)</dt>
+
+<dt>[Step 3: Submitting a GitHub Pull Request](step-3)</dt>
```
mcdurdin, line 24: This should not be a definition list because it only contains terms. Instead, just make it a bullet list `* step 0` ...

mcdurdin (changes requested):
I don't have permission to update the Markdown parser here so we will add it as a new PR and we can rebase this PR once it lands.

(Note, the version of Parsedown that is on keymanapp/keyman.com#192 is the one that we will use here).

DavidLRowe:
Sorry to have left this sitting so long. It dropped off my radar. I went to look for it online and realised that it hadn't been merged because more work was needed. 

I've attempted to address the problems you both (@darcywong00 and @mcdurdin) have pointed out. I used "xyz" rather than "keyboard" in angle brackets to avoid display issues.

Ignore my comment about list text being repeated (in the "Update step-0.md" commit. What I was seeing was GitHub showing me deletions and additions when I previewed the changes.

In Step 2, I'm not really happy with the "One keyboard per folder" heading. I think the text has changed significantly and the header should probably change too. Any suggestions?!

mcdurdin (approved):
Looks good to me!

> Sorry to have left this sitting so long. It dropped off my radar. I went to look for it online and realised that it hadn't been merged because more work was needed.

No problems -- understand having many things to look after 😉 

> In Step 2, I'm not really happy with the "One keyboard per folder" heading. I think the text has changed significantly and the header should probably change too. Any suggestions?!

I'm not really coming up with any better ideas just now.

DavidLRowe:
@mcdurdin Thanks for catching the one I missed. I think it's ready to merge now.

mcdurdin:
Thank you @DavidLRowe for this excellent PR 🥇  I always get excited when we get documentation improvements -- we don't do enough of them I think... 😄

PR #255, opened 2020-11-16
chore: update parsedown

mcdurdin:
This adds support for latest version of Parsedown to master branch and adds ParsedownExtra for support for additional features such as definition lists.

Will be used for content such as #242.

Throws in vscode styling while we are at it.

Note that this diverges from the work in staging branch (e.g. #256) because we do not use Composer yet in the master branch.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,7 @@
+{
+  "workbench.colorCustomizations": {
+    "activityBar.background": "#4A1E3B",
+    "titleBar.activeBackground": "#672A53",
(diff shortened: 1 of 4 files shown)


darcywong00:
> Note that this diverges from the work in staging branch (e.g. #256) because we do not use Composer yet in the master branch.

How will the merge look at the end of sprint from master->stable?

mcdurdin:
> How will the merge look at the end of sprint from master->stable?

It might need a little conflict cleanup in mdhost.php but not too much.

darcywong00 (approved):
lgtm

PR #256, opened 2020-11-16
chore: sync common files

mcdurdin:
This synchronises the new `MarkdownHost` class between keymanapp/keyman.com#192 and help.keyman.com.

It adds also `ParsedownExtra` support so we can do definition lists, which I have validated with the `contextExOutput` file.


_includes/md/mdhost.php
```diff
@@ -1,57 +1,16 @@
 <?php
   require_once('includes/template.php');
```
darcywong00, line 2: The keyman.com version also has this:
```
require_once __DIR__ . '/../../autoload.php';
```

darcywong00 (approved):
lgtm  (is autoload not needed?)

_includes/md/mdhost.php
mcdurdin: Yes, on help.keyman.com, the autoload is included automatically, but we haven't yet done the work on keyman.com for that.

darcywong00 (changes requested):
oops, I noticed the check fails from PHP errors referencing Parsedown.php
history_utils.php
knowledge-base/index.php

darcywong00 (approved):
lgtm

PR #258, opened 2020-11-24
Move howto pages

ermshiperete:
- "Using Keyman for Linux with KDE" from community site
- "How to install IBus and Keyman on Arch Linux and derivatives" and "How to install nightly packages of Keyman for Linux" from Keyman wiki

Once this PR is merged we should replace the content of the original pages with links to the KB articles

### DIFF
--- a/knowledge-base/kb0101.md
+++ b/knowledge-base/kb0101.md
@@ -1,0 +1,36 @@
+# Using Keyman for Linux with KDE
+
+Unfortunately when using the KDE desktop environment the support for ibus (which Keyman for Linux uses underneath) isn’t that great. If you want to use Keyman with KDE you’ll have to do some manual steps:
+
+1. Install additional packages: in a terminal window, run:
+
+    ```bash
+    sudo apt update
+    sudo apt install ibus-qt4 ibus-gtk
+    ```
+
+2. Add
+
+    ```bash
+    export GTK_IM_MODULE=ibus
+    export XMODIFIERS=@im=ibus
+    export QT_IM_MODULE=ibus
+    ```
+
+    to your `$HOME/.xsessionrc`.
+
+3. Auto-start the ibus daemon on every login. Run the following command in a terminal window:
+
+    ```bash
+    echo "ibus-daemon -d -x -r -n kde" > $HOME/.config/autostart-scripts/ibus-daemon-autostart.sh
+    chmod +x $HOME/.config/autostart-scripts/ibus-daemon-autostart.sh
+    ```
+
+4. Reboot so that the changes you made get applied to the new session.
+
+Now you should be able to install Keyman for Linux and add and use Keyman keyboards.
+
+**NOTE:** it seems that some KDE specific programs like Konsole aren’t designed to work with
+ibus/Keyman keyboards. To test if a Keyman keyboard works you can e.g. test with LibreOffice Writer.
+
+(cf. <https://wiki.debian.org/I18n/ibus>)
--- a/knowledge-base/kb0101.md
+++ b/knowledge-base/kb0101.md
@@ -1,6 +1,8 @@
-# Using Keyman for Linux with KDE
+# HOWTO: Using Keyman for Linux with KDE
 
-Unfortunately when using the KDE desktop environment the support for ibus (which Keyman for Linux uses underneath) isn’t that great. If you want to use Keyman with KDE you’ll have to do some manual steps:
+Unfortunately when using the KDE desktop environment, the support for ibus (which Keyman for Linux
+uses underneath) isn’t that great. If you want to use Keyman with KDE you’ll have to do some manual
+steps:
 
 1. Install additional packages: in a terminal window, run:
 
--- a/knowledge-base/kb0101.md
+++ b/knowledge-base/kb0101.md
@@ -30,7 +30,12 @@ Unfortunately when using the KDE desktop environment the support for ibus (which
 
 Now you should be able to install Keyman for Linux and add and use Keyman keyboards.
 
+---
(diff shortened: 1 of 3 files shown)


knowledge-base/kb0103.md
```diff
@@ -0,0 +1,32 @@
...
+sudo echo have sudo
+```
+
+Next paste these commands:
+
+```bash
+wget -qO - http://linux.lsdev.sil.org/downloads/sil-testing.gpg | sudo apt-key add -
+sudo add-apt-repository "deb http://linux.lsdev.sil.org/ubuntu $(lsb_release -sc)-experimental main"
+sudo apt-get update
+sudo apt-get install keyman
```
darcywong00, line 28: Does `onboard` get installed with this?

knowledge-base/kb0103.md
```diff
@@ -0,0 +1,32 @@
+# How to install nightly packages of Keyman for Linux
+
+Nightly alpha packages of Keyman for Linux are available on llso.
```
darcywong00, line 3: You use `LLSO` later in the doc. What is  `LLSO` an abbreviation for?

knowledge-base/kb0103.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Warning
+
+**WARNING:** "nightly" experimental packages are untested, unsupported, and may cause problems on
```
darcywong00, line 7: What happens if the user also added the keyman-nightly or keyman PPA?
Which one would they get with `sudo apt-get install keyman`?

knowledge-base/kb0101.md
```diff
@@ -0,0 +1,41 @@
+# Using Keyman for Linux with KDE
+
+Unfortunately when using the KDE desktop environment the support for ibus (which Keyman for Linux uses underneath) isn’t that great. If you want to use Keyman with KDE you’ll have to do some manual steps:
```
darcywong00, line 3:
```suggestion
Unfortunately when using the KDE desktop environment, the support for ibus (which Keyman for Linux uses underneath) isn’t that great. If you want to use Keyman with KDE you’ll have to do some manual steps:
```

knowledge-base/kb0101.md
```diff
@@ -0,0 +1,41 @@
+# Using Keyman for Linux with KDE
```
darcywong00, line 1: I don't know if it's a convention, but I notice some other knowledge base articles prefix the title
```
# HOWTO: Using Keyman for Linux with KDE
```

Applies to the 3 md files on this PR.

knowledge-base/kb0102.md
```diff
@@ -0,0 +1,72 @@
...
+(Packages `kmflcomp`, `libkmfl`, `ibus-kmfl` belong to an old version of Keyman, ignore them.)
+
+Several dependencies of Keyman are not included into KOSMOS system, install them also from `pamac`:
+
+* python-numpy (Repositories)
+* python-magic (AUR)
+* python-requests-cache (AUR)
+
+Now all Keyman's utilities should work. Search Application Launcher for "Keyman Keyboards"
+(km-config) for installing required keyboards. Layout will be automaticly added into IBus, if not —
```
darcywong00, line 67: sp "automatically"

knowledge-base/kb0102.md
```diff
@@ -0,0 +1,72 @@
...
+makepkg
+makepkg -i
+```
+
+Run `pamac` from the Application Launcher (it is called "Add/Remove Software") or from console
+(`pamac-manager`). Go to the Preference menu (button with 3 vertical dots in the upper right corner)
+and activate "Enable AUR support". Return to the main window and search for "keyman", there should
+be two applications, install them:
+
+* keyman 11.0.122-1
```
darcywong00, line 55: Do we know if that's still version 11.0?

mcdurdin (approved):

knowledge-base/kb0103.md
ermshiperete: It depends on the settings if recommendations get also installed.

Should we add it to the command here?

knowledge-base/kb0103.md
mcdurdin: Can we make sure that we use the full name and instructions on how to use it -- external readers won't be able to understand llso or what to do with it.

knowledge-base/kb0102.md
mcdurdin: We really should be using the latest version and not recommending people install the old version. In fact, it'd be better not to reference a version number at all -- it's our job to make sure it works in later versions too!

knowledge-base/kb0103.md
ermshiperete: Is my updated version acceptable?

knowledge-base/kb0103.md
darcywong00: I think it would be good to include `onboard` here.

I'm thinking of the user on Groovy who's keen to get `onboard`.

knowledge-base/kb0102.md
ermshiperete: AFAIKT they won't see the old version, just the current one. IMO we should leave the old version number in there. I don't want to add Arch Linux (and possibly derivatives like Kosmos) to our list of environments we have to test every time we release. If we leave the old version number it's clear that this was working for version 11 and that it might have to be slightly adjusted for a newer version. Unfortunately Linux users are more or less used to this if they try to use apps that are not included in the official repos of their distro, or if they use a distro for advanced users.

knowledge-base/kb0103.md
ermshiperete: Done

darcywong00 (approved):
lgtm

mcdurdin (approved):

PR #263, opened 2020-11-29
chore:  master staging

mcdurdin:


_includes/includes/ext/parsedown/ParsedownExtra.php
```diff
@@ -0,0 +1,538 @@
+<?php
```
darcywong00, line 1: hm, I thought the staging site gets Parsedown and ParsedownExtra from composer.

mcdurdin: Yep. We added ParsedownExtra to master but need to take it out again from staging... :( I will update this PR

darcywong00 (approved):
lgtm

PR #268, opened 2020-12-04
feat(linux): Rename onboard package

ermshiperete:
Update the documentation for the renamed onboard package. The package name is now `onboard-keyman` instead of `onboard`.
Part of keymanapp/keyman#3966.


darcywong00 (approved):
lgtm

PR #299, opened 2021-01-24
chore: hosts and header maintenance

mcdurdin:
* Harmonize KeymanHosts.php with keyman.com
* Exclude staging site from search engines


darcywong00 (approved):
lgtm

PR #333, opened 2021-03-05
docs: Add whatsnew to Keyman Engine for Android

darcywong00:
* Move breaking Keyman Engine for Android info from keymanapp/keyman#4610 to here
* Also add Python and Groovy to prism


mcdurdin (approved):

mcdurdin:
LGTM, but can we avoid using `KMEA` and instead use `Keyman Engine for Android` or just `Keyman Engine`?

PR #348, opened 2021-03-12
feat: add virtual key code visual chart

mcdurdin:
Adds visual documentation of the virtual key layout:

image

*** note: updated graphic because original screenshot I had here was of an older iteration which had a typo.

Also:
* Two ISO codes were interchanged
* Adds `B11` for ABNT2 keyboard.

### DIFF
--- a/developer/language/guide/virtual-keys.md
+++ b/developer/language/guide/virtual-keys.md
@@ -179,7 +179,7 @@ The following table lists all of the common virtual key codes:
 
 | Key Code    | ISO9995 | Key Name |     | Key Code    | ISO9995 | Key Name                |
 |-------------|---------|----------|-----|-------------|---------|-------------------------|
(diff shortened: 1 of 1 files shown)


jahorton (commented):
So B10 and B11 have the same default keycaps?

mcdurdin:
> So B10 and B11 have the same default keycaps?

Well, on a Brazilian keyboard, `B10` has a different key cap (`; :`) but I feel like it's going to get too complicated to represent that here... Similarly, `B00` and `D13` do not have the same key caps on non-US keyboards.

jahorton:
> Well, on a Brazilian keyboard, `B10` has a different key cap (`; :`) but I feel like it's going to get too complicated to represent that here... Similarly, `B00` and `D13` do not have the same key caps on non-US keyboards.

Ah... I didn't even notice the other case.  This one is more prominent due to the obvious neighbor key.

jahorton (approved):

PR #349, opened 2021-03-12
fix: markdown header parser was not parsing some headers

mcdurdin:
This replaces the somewhat fragile regex with a simple line-based parser of the headers. Seems to be somewhat cleaner.


darcywong00 (approved):
lgtm

PR #364, opened 2021-03-24
chore: Revision of the iOS 14.0 engine guide

jahorton:
Parallels https://github.com/keymanapp/keyman/pull/4772.


developer/engine/iphone-and-ipad/14.0/guides/system-keyboard/index.php
```diff
@@ -75,10 +76,35 @@ class="asset  asset-image at-xid-6a00d83432368053ef01b7c7604a62970b image-full i
...
-  will probably want to make that a lot prettier and include detailed instructions for users on how to enable the system keyboard
-on their own.  (Unfortunately, apps cannot directly summon the appropriate menu.)</p>
+  will probably want to make that a lot prettier.
+<ol>
+  <li>For those who are technically savvy, you may wish to refer to 
+  <a href="https://github.com/keymanapp/keyman/blob/stable-14.0/ios/keyman/Keyman/SWKeyboard/ImageBannerViewController.swift">Keyman's implementation</a>
+   of the banner image, which adjusts to the user's device.</li>
+</ol>
+</p>
+-->
```
jahorton, line 96: Of possible interest.

Is the detail mentioned here worth turning into an issue on the main repo?  Or, should we wait for that eventual API rework (or just the next version) before we talk about making an API for "show / don't show branding"?

darcywong00: imho, I'd leave all the branding verbage out

darcywong00 (approved):
lgtm

developer/engine/iphone-and-ipad/14.0/guides/system-keyboard/index.php
jahorton: That's why it's `<!--`commented out, HTML-style`-->`.

On a side note, I had to use code formatting on the... tags?... so that the text between them wasn't hidden in the actual comment.  Huh.

PR #371, opened 2021-04-01
chore: A15S1 staging-branch update

jahorton:
Since we _just_ released v14, a lot of our recent help work was applied to `master`, rather than `staging`.  It's worth noting that some of the 'alpha' versions of our apps (like for iOS) will point to `staging`, so we should probably update `staging` with the new stuff.


darcywong00 (approved):
lgtm

mcdurdin:
This needs to be done with a separate branch for tracking purposes. Closing this PR.

jahorton:
Please tag this PR when you do so.

mcdurdin:
> Or are you saying that simply making a temp branch off of master, thus capturing a 'snapshot' of it, is sufficient?

Yes

PR #413, opened 2021-06-09
chore: Add link to YouTube channel

darcywong00:
@MakaraSok has been creating content on the Keyman YouTube channel.

Rather than updating help.keyman.com each time, this just adds a link to the channel.

If there's a better place on the site than developer/videos, I'm open to it.


darcywong00:
@mcdurdin suggests:
> I think it's a good idea to add links to the videos as well, because that makes them much more visible in the help. In fact, a link from the relevant help page (e.g. if a video talks about Keyman for Windows Configuration, then a link from the Config help) is also good.


PR back in draft until I get around to adding more links

darcywong00:
https://github.com/keymanapp/help.keyman.com/pull/413/commits/4c845d55b3fa65258fb7e805fa31c7d6fcb912ae adds a video link of Keyman Developer features on the Developer intro page

mcdurdin (approved):

PR #420, opened 2021-06-17
docs: Add KeymanWeb option flags useAlerts and setActiveOnRegister

darcywong00:
This documents the added options flag `useAlerts` from keymanapp/keyman#5302. 
Also adds the missing option `setActiveOnRegister`

Most of the changes are from seeding the 15.0 KeymanWeb engine content, so I suggest reviewing the 3rd commit onward.


mcdurdin:
Can we please split the seed and the changes into two separate PRs?

mcdurdin (approved):
LKGTM

K = kinda.

Because `'true'` and `'false'` are abominations!

PR #441, opened 2021-06-29
feat: Adjust serving .md files when running with Docker

ermshiperete:


mcdurdin (approved):

mcdurdin:
I haven't attempted to validate the .htaccess changes, but as they are only for your own internal use I think that I will leave any fixes up to you 😁

PR #455, opened 2021-08-25
chore: Add kb article on installing for M1 Mac

darcywong00:
This adds a knowledge base article on installing Keyman for macOS on an M1.

Reference
https://community.software.sil.org/t/does-keyman-support-m1-mac/5213


mcdurdin (approved):
Can we add a note that this is a temp fix until we release an M1 update with 15.0?

PR #460, opened 2021-09-24
fix: logo overlap on some sizes

### LINKED ISSUE #261: bug: Keyman logo overlaps with a menu when page down
**Describe the bug**
In Keyman Configuration, go to Support and then scroll down to see more context. You may see on the top right that the Keyman logo overlaps with a menu.

image

**Expected behavior**
The logo should be where it is, but the menu should be scootered to the right to avoid overlapping with the logo.

---

**Keyman for macOS:**
 - OS: macOS Mojave, 
 - Keyman version: 14.0.190
 - Target Application: Keyman Configuration
 
---

**Additional context**
This similar behavior is seen when browsing [a help page](https://help.keyman.com/?_ga=2.38951403.977768928.1606093069-30305726.1574652282) on a narrower window in Chrome on Windows 10 Pro:
image

mcdurdin:
Fixes #261. Easiest way to fix was to remove the 600px stop which was inconsistent in its styling in any case.

### DIFF
--- a/cdn/dev/css/template.css
+++ b/cdn/dev/css/template.css
@@ -1341,7 +1341,7 @@ body[data-device='Unknown'] #osk-tablet-container {
  * TABLET CSS
  * ***********************************/
 
-@media only screen and (min-width: 768px) and (max-width: 1024px){
+@media only screen and (min-width: 600px) and (max-width: 1024px){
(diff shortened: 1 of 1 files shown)


darcywong00 (approved):
lgtm

PR #464, opened 2021-09-27
auto: Keyman for windows help deployment

mcdurdin:


mcdurdin:
Testing; this one is still buggy!

PR #477, opened 2021-09-30
docs: if, set and system stores

### LINKED ISSUE #366: docs: if, set, and Using variable stores topics do not document use of system stores
* They need to document use of `&layer` etc.
* Some of the 'related topics' say 'Variable system stores' but should say 'Variable stores'
* They need to note that `if()` must be at start of rule (see also https://help.keyman.com/developer/language/guide/rules):

```
Certain statements must appear in specific positions in the rule. The following order is enforced by the compiler:

* nul
* if(), baselayout() and platform()
* All other statements and characters
```

mcdurdin:
Fixes #366.

Note: I have not attempted to add any additional content to the guide/variable-stores topic, as it doesn't really make sense to document the system store modes for `if()` and `set()` there.


developer/language/reference/if.md
```diff
@@ -0,0 +1,127 @@
...
+
+```
+if(&platform = 'web') + any(letterGroup1) > index(outputGroup1, 2)
+```
+
+## Platforms
+
+The `if()` statement can be used in keyboards on the following platforms.
+
+Currently, only Windows supports the options.htm form to set variable stores.
```
darcywong00, line 99: Keyman for Linux supports options dialog in keymanapp/keyman#2566

Are variable stores stored differently?

mcdurdin: > Are variable stores stored differently?

I'm not sure I understand the question?

darcywong00: I know on Linux the options.htm form can toggle and save keyboard options.

I was checking if variable stores are stored the same way.

mcdurdin: variable stores === keyboard options :)

darcywong00 (approved):
lgtm

PR #501, opened 2022-02-09
chore: Update kb0104.md PostEvent to All for tccutil

mcdurdin:
Change from PostEvent to All for tccutil, based on experience in the team.


darcywong00 (approved):
lgtm

PR #509, opened 2022-03-02
docs(developer): Keyman Developer Server

mcdurdin:
Documentation per checklist item in keymanapp/keyman#6178


rc-swag (approved):
lgtm

DavidLRowe:
Wow, I wish I'd had all the information in server.md years ago! It looks really good. 

Is it worth mentioning (if it's still the case) that "Open in browser" uses your default browser and if it's not Chrome (or Chromium based) the results may not be satisfactory?

mcdurdin:
> Wow, I wish I'd had all the information in server.md years ago! It looks really good.

Well, most of this is new from December 2021 😀 



> Is it worth mentioning (if it's still the case) that "Open in browser" uses your default browser and if it's not Chrome (or Chromium based) the results may not be satisfactory?

Any modern browser should be okay -- doesn't need to be Chrome based as of Keyman Developer 15! IE and legacy Edge don't cut the mustard, but Chrome, Firefox, Modern Edge (Chromium-based), even probably Opera (though that is untested) should be fine.

DavidLRowe:
LGTM

PR #519, opened 2022-03-18
docs: KB0108 - No such text input source

mcdurdin:
Ref: https://github.com/keymanapp/keyman/issues/5578
Ref: https://community.software.sil.org/t/keyman-will-not-install/5759/9


rc-swag (approved):
lgtm

PR #523, opened 2022-03-25
docs: updated info on seralised input behaviour

rc-swag:
From version 15 the seralized input pipline will not be active
if a Keyman keyboard is not active. Update the document to reflect this.

### DIFF
--- a/knowledge-base/kb0097.md
+++ b/knowledge-base/kb0097.md
@@ -2,7 +2,9 @@
 
 ## Background
 
-When Keyman is running, it controls the keyboard input queue in Windows in order to guarantee the sequence of events. You should not normally need to change this behaviour; however it is possible to switch off the control of the keyboard input queue if you suspect a conflict with another utility. From version 15 if a Keyman keyboard is not active, keyman will disable the serialised input. If you are running an older version of Keyman upgrade to the latest version. If that does not resolve your issue then try disabling the serialised input as described below.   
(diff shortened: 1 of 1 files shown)


knowledge-base/kb0097.md
```diff
@@ -2,7 +2,8 @@
 
 ## Background
 
-When Keyman is running, it controls the keyboard input queue in Windows in order to guarantee the sequence of events. You should not normally need to change this behaviour; however it is possible to switch off the control of the keyboard input queue if you suspect a conflict with another utility. [Learn more...](https://blog.keyman.com/2018/10/the-keyman-keyboard-input-pipeline/)
+When Keyman is running, it controls the keyboard input queue in Windows in order to guarantee the sequence of events. You should not normally need to change this behaviour; however it is possible to switch off the control of the keyboard input queue if you suspect a conflict with another utility. From version 15 if a Keyman keyboard is not active, keyman will disable the serialised input. If you are running an older version of Keyman upgrade to the latest version. If that does not resolve your issue then try disabling the serialised input as described below.
```
mcdurdin, line 5:
```suggestion
When Keyman is running, it controls the keyboard input queue in Windows in order to guarantee the sequence of events. You should not normally need to change this behaviour; however it is possible to switch off the control of the keyboard input queue if you suspect a conflict with another utility.

From version 15, Keyman will enable serialised input only if a Keyman keyboard is active. If you are running an older version of Keyman, you should upgrade to the latest version. If that does not resolve your issue then try disabling the serialised input as described below.
```

mcdurdin (approved):
Minor tweak to intro para :grin:

PR #526, opened 2022-04-03
fix: help offset

mcdurdin:
Relates to keymanapp/keyman#2750.

When you load a page with a `#toc-` hash, it would not scroll to the correct place in the document, because the `#toc-` anchors are all defined dynamically after page load.

This adds a slightly hacky fix to scroll the anchor into view, leveraging the existing toc elements to access the custom scroll code.

Note that this jquery code is all very old, but changing that is a much more significant update which needs site-wide consideration.


darcywong00 (approved):
lgtm

PR #543, opened 2022-04-13
docs: clarify setup.exe must be unsigned

mcdurdin:
This applies only to building self-extracting installers...


darcywong00 (approved):
lgtm

PR #547, opened 2022-04-26
fix: Formatting on engine/android in-app guide

darcywong00:
The sample code for adding a dictionary didn't have the escape special characters for PHP.

* Fixed the format in the 14.0 guide
* Went ahead and migrated the 15.0 guide to Markdown (for maintenance)


jahorton:
I know this probably isn't the best place for this, but a little something I ran into when attempting to 'test' this PR:

While my Mac setup is sufficient for testing the current `staging` branch, there's something different for `master` that causes a "Internal Server Error" on my `localhost` Apache-based setup.

rc-swag (approved):
lgtm

PR #548, opened 2022-04-26
feat: adds new properties for Web's getKeyboard API

jahorton:
Documents the features added by https://github.com/keymanapp/keyman/pull/6567.

image

### DIFF
--- a/developer/engine/web/14.0/reference/core/getKeyboard.php
+++ b/developer/engine/web/14.0/reference/core/getKeyboard.php
@@ -37,7 +37,7 @@ obj_spec('keyboard',
   "Font", "optional string", "The font packaged with the keyboard to support its use.",
   "OskFont", "optional string", "The font packaged with the keyboard to properly display specialized OSK characters.",
   "HasLoaded", "boolean", "<code>true</code> if the keyboard has been fully loaded.  <code>false</code> otherwise.",
-  "IsRTL", "optional boolean", "<code>null</code> whenever HasLoaded is <code>false</code>.  Otherwise, indicates whether or not the keyboard's target script is RTL."
+  "IsRTL", "optional boolean", "<code>null</code> whenever HasLoaded is <code>false</code>.  Otherwise, <code>true</code> if the keyboard's target script is written right-to-left."
   );
 ?>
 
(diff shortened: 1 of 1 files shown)


darcywong00:
I think in previous sprints, @mcdurdin said we'll merge help PR's to `master` instead of `staging`. When needed, we'll just update the staging branch.

I did see your comment in the other PR about not being able to view help.keyman on the master branch though 😞

darcywong00:
Other than running composer install, the other recent change I can think of is

https://github.com/keymanapp/keyman/blob/master/resources/devbox/iis-https/README.md

jahorton:
> Other than running composer install, the other recent change I can think of is
> 
> https://github.com/keymanapp/keyman/blob/master/resources/devbox/iis-https/README.md

Well, that part doesn't exactly fly on Macs - they don't exactly do IIS.

mcdurdin:
> Well, that part doesn't exactly fly on Macs - they don't exactly do IIS.

Running the websites on Macs has never been fully supported.

Yes, please work against `master` -- we are not deploying to `staging` for these changes. (We tend to only use `staging` when we are making substantial site-wide changes that are tied to the release and won't work otherwise, such as when we were doing the new keyboard search.)

developer/engine/web/14.0/reference/core/getKeyboard.php
```diff
@@ -35,7 +35,9 @@
   "CountryCode", "optional string", "A three-letter code corresponding to the country.",
   "KeyboardID", "optional string", "<em>Deprecated.</em>  A unique identifier for the keyboard.  (Use 'InternalName' instead.)",
   "Font", "optional string", "The font packaged with the keyboard to support its use.",
-  "OskFont", "optional string", "The font packaged with the keyboard to properly display specialized OSK characters."
+  "OskFont", "optional string", "The font packaged with the keyboard to properly display specialized OSK characters.",
+  "HasLoaded", "boolean", "<code>true</code> if the keyboard has been fully loaded.  <code>false</code> otherwise.",
+  "IsRTL", "optional boolean", "<code>null</code> whenever HasLoaded is <code>false</code>.  Otherwise, indicates whether or not the keyboard's target script is RTL."
```
mcdurdin, line 40:
```suggestion
  "IsRTL", "optional boolean", "<code>null</code> whenever HasLoaded is <code>false</code>.  Otherwise, <code>true</code> if the keyboard's target script is written right-to-left."
```

mcdurdin (approved):

PR #565, opened 2022-06-28
fix: use version sort for keyboard help redirection

### LINKED ISSUE #564: bug: keyboard help forwarding is getting wrong version for fv_tlingit
https://help.keyman.com/keyboard/fv_tlingit is forwarding to 
https://help.keyman.com/keyboard/fv_tlingit/9.2/fv_tlingit instead of https://help.keyman.com/keyboard/fv_tlingit/10.0/fv_tlingit. Suspect that it is doing a string rather than numerical comparison (string "10.0" < "9.2" but numeric 10.0 > 9.2).

mcdurdin:
Fixes #564.

### DIFF
--- a/keyboard/redir.php
+++ b/keyboard/redir.php
@@ -13,7 +13,7 @@
     $v = "$q/{$_REQUEST['v']}";
     if(!file_exists("$v/$q.php")) {
       unset($v);
-  }
+    }
   }
   if(!isset($v)) {
     $versions = array_filter(glob($q . '/*'), 'is_dir');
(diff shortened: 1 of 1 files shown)


jahorton (approved):

keyboard/redir.php
```diff
@@ -13,7 +13,7 @@
     $v = "$q/{$_REQUEST['v']}";
     if(!file_exists("$v/$q.php")) {
       unset($v);
-    }
+  }
```
jahorton, line 16: Maybe undo this removal of indentation?

```suggestion
    }
```

PR #566, opened 2022-06-28
docs: Update casing-support.md to clarify touch layout behaviour only

mcdurdin:
(See keymanapp/keyman#6799 discussion also.)

`set(&layer)` is [documented](https://help.keyman.com/developer/language/reference/layer) as impacting only touch keyboards:

> The `&layer` system variable store can be used to set or test the current layer on the touch keyboard.

But current behaviour has it impacting the desktop OSK as well. This does not seem right to me... 

So this is a documentation workaround this with additional hoops for keyboard authors. I am not sure that we should be doing this?


mcdurdin:
> I am not sure that we should be doing this?

Nope, we shouldn't. New PRs on their way.

mcdurdin:
Replaced by #569.

PR #586, opened 2022-08-09
docs: first cut auto convert html to md

rc-swag:
Convert the core api docs to markdown.


developer/core/desktop/16.0/c-api/state-api.md
```diff
@@ -0,0 +1,594 @@
...
+`num_items`
+
+A pointer to a result variable: The number of items in the action item list including the [`KM_KBP_IT_END`](#KM_KBP_IT_END) terminator. May be null if that information is not required.
+
+Returns
+-------
+
+A pointer to a [`km_kbp_action_item`](#km_kbp_action_item) array, of `*``num_items` in length. This data becomes invalid when the state object is destroyed, or after a call to [`km_kbp_process_event()`](#km_kbp_process_event). Do not modify the contents of this array. The returned array is terminated with a [`KM_KBP_IT_END`](#KM_KBP_IT_END) entry.
+
+km\_kbp\_state\_to\_json()
```
darcywong00, line 292: The link-checker during the CI build failed because the page
> Getting links from: http://localhost:8888/developer/core/desktop/16.0/json-schema
>├─BROKEN─ http://localhost:8888/developer/core/desktop/16.0/c-api/state-api.php#km_kpb_state_to_json (HTTP_404)

So the  json-schema page will need to be updated to link to this section. (I don't remember the details for setting up a Markdown anchor)

darcywong00: Ah, it's something like
```
/developer/core/desktop/16.0/c-api/state-api#toc-km-kbp-state-to-json-
```

developer/core/desktop/16.0/c-api/state-api.md
```diff
@@ -0,0 +1,594 @@
...
+============================
+
+Description
+-----------
+
+These provide the results of processing a key event to the Platform layer and should be processed by the Platform layer to issue commands to the OS text services framework to transform the text store in the Client Application, among other actions.
+
+Specification
+-------------
+```
```
darcywong00, line 17:
```suggestion
```c
```
This will apply C-syntax highlighting for the code blocks. (Probably applies to all the code blocks on this PR)

rc-swag:
Internal Links 
Using the PHP  Markdown Extra syntax defined [here](https://michelf.ca/projects/php-markdown/extra/#spe-attr). The Headers now have an id set that could then be linked back to. 
```
Header 1            {#header1}
========
```
Then you can create links to different parts of the same document like this:
```
[Link back to header 1](#header1)
```
Things to note I was able to use this for the majority of the links and they were already linking to "Headers".
For Non-headers for example `KM_KBP_IT_END`  or `KM_KBP_OPT_KEYBOARD` The nearest header of `Values` worked well as this was the section that had all the definitions.
Finally, there were some links that just linked a few lines up to the definition of function parameters suchas `out`, `buf` and `state`. These were just completely removed as they did not really achieve much as what they where linking to in most cases was only a few lines above the link.

developer/core/desktop/16.0/c-api/context-api.md
```diff
@@ -0,0 +1,548 @@
+---
+title: Context
+---
+
+The context is the text prior to the insertion point (caret, cursor). The context is constructed by the Platform layer, typically by interrogating the Client Application. The context will be updated by the engine for keystroke events. If the Platform layer code caches the context, the context should be reset when a context state change is detected. Context state changes can occur when the user uses the mouse to move the insertion point, uses cursor keys, switches applications or input fields, or presses hotkeys such as Ctrl+N to start a new document. The full set of context state change triggers is up to the Platform layer.
```
mcdurdin, line 5: Can we wrap all long lines? It makes future review much easier.

I use an extension in VS Code called [Rewrap](https://marketplace.visualstudio.com/items?itemName=stkb.rewrap) which does this pretty well

developer/core/desktop/16.0/c-api/context-api.md
```diff
@@ -0,0 +1,548 @@
+---
+title: Context
```
mcdurdin, line 2: This probably needs to be a little longer, e.g. `Context - Keyman Core API` (same applies for titles in all files)

developer/core/desktop/16.0/c-api/context-api.md
```diff
@@ -0,0 +1,548 @@
...
+---
+
+The context is the text prior to the insertion point (caret, cursor). The context is constructed by the Platform layer, typically by interrogating the Client Application. The context will be updated by the engine for keystroke events. If the Platform layer code caches the context, the context should be reset when a context state change is detected. Context state changes can occur when the user uses the mouse to move the insertion point, uses cursor keys, switches applications or input fields, or presses hotkeys such as Ctrl+N to start a new document. The full set of context state change triggers is up to the Platform layer.
+
+Context can also contain positional Markers (also known as 'deadkeys' in kmn keyboards), which are transitory state flags that are erased whenever a context state change is detected. Markers are always controlled by the Engine.
+
+Contexts are always owned by their state. They may be set to a list of context\_items or interrogated for their current list of context items.
+
+km\_kbp\_context\_type enum  {#km_kbp_context_type}
+===========================
```
mcdurdin, line 12: I think I'd prefer to use the other form of headings for consistency with other pages on the site, and tweak the formatting slightly at the same time:

```suggestion
# `km_kbp_context_type` enum  {#km_kbp_context_type}
```

developer/core/desktop/16.0/c-api/context-api.md
```diff
@@ -0,0 +1,548 @@
...
+
+Description
+-----------
```
mcdurdin, line 15: Same thing re headings here and elsewhere in these docs:

```suggestion
## Description
```

developer/core/desktop/16.0/c-api/context-api.md
```diff
@@ -0,0 +1,548 @@
...
+Specification
+-------------
+```c
+    enum km_kbp_context_type {
+      KM_KBP_CT_END,
+      KM_KBP_CT_CHAR,
+      KM_KBP_CT_MARKER
+    };
+```
+Values {#values}
```
mcdurdin, line 28: This is a very generic link ref (`#values`). If we want to use it (I don't think we do), it should be `#km_kbp_context_type-values` or similar

developer/core/desktop/16.0/c-api/context-api.md
```diff
@@ -0,0 +1,548 @@
...
+
+A [`km_kbp_context_item`](#km_kbp_context_item) of this type marks the end of an array of context items.
+
+`KM_KBP_CT_CHAR`
+
+The context item contains a Unicode Scalar Value which must be accessed through the [`km_kbp_context_item.character`](#km_kbp_context_item) union member.
+
+`KM_KBP_CT_MARKER`
+
+The context item contains a positional marker which must be accessed through the [`km_kbp_context_item.marker`](#km_kbp_context_item) union member.
```
mcdurdin, line 41:
```suggestion
* `KM_KBP_CT_END`
: A [`km_kbp_context_item`](#km_kbp_context_item) of this type marks the end of an array of context items.

* `KM_KBP_CT_CHAR`
: The context item contains a Unicode Scalar Value which must be accessed through the [`km_kbp_context_item.character`](#km_kbp_context_item) union member.

* `KM_KBP_CT_MARKER`
: The context item contains a positional marker which must be accessed through the [`km_kbp_context_item.marker`](#km_kbp_context_item) union member.
```

We can use [Definition Lists](https://michelf.ca/projects/php-markdown/extra/#def-list) to improve formatting here.

developer/core/desktop/16.0/c-api/index.md
```diff
@@ -0,0 +1,281 @@
+---
+title: C API Reference
```
mcdurdin, line 2:
```suggestion
title: Reference - Keyman Core API
```

mcdurdin (changes requested):
This is looking good. I've requested some changes to the style of markdown used, particularly for headings and definition lists. But those changes are not critical.

developer/core/desktop/16.0/c-api/context-api.md
rc-swag: Yes will change the name as it could avoid a duplicate clash also, but we do already use it as mentioned [here](https://github.com/keymanapp/help.keyman.com/pull/586#issuecomment-1480569195) all the previous links in the php version that went direct to the types `KM_KBP_CT_???` now have links the heading above as it is sufficient and works with markdown.

developer/core/desktop/16.0/c-api/context-api.md
rc-swag: It is good in isolation, although the left-hand tree view looks a little redundant now. 
image

rc-swag:
I change then reverted the
 ```# km\_kbp\_action\_item struct {#km_kbp_action_item} ``` to 
# `km_kbp_action_item` struct {#km_kbp_action_item}```

As the backtick overides the heading in markdown.  So escaping the with \ needs to stay.

I added the lists but they made no difference the rendering visually, even though the html now has the <dl > tags.

All the heading are updated to the same style used in other markdown documents on the site.

mcdurdin (approved):

PR #588, opened 2022-08-16
chore: :axe:  Migrate help site from IIS to Apache

darcywong00:
For the help.keyman.com portion of #keymanapp/keyman.com#337

This is the initial pass of converting from web.config to .htaccess files so the help.keyman.com site can run on Linux.

* I left `php_value` intact in the root .htaccess file since that will change on the developer/deployment.

I also haven't setup SSL locally so my dev also has the following commented out:
https://github.com/keymanapp/help.keyman.com/blob/64b174b0f2b1fc5eb9485809994259980f2d5dbe/.htaccess#L8-L13

@jahorton and @ermshiperete have previously run their local site on non-Windows, so maybe they can spot check.

----
I'll be verifying the links with 
```
composer check-links
```

PR #594, opened 2022-08-29
fix: :axe: Don't upper-case breadcrumb links

darcywong00:
Part of the :axe: chain for #588 

Since filenames on Linux are case-sensitive, this removes upper-casing bread crumb links.

### Sample Screenshots of Updated Links
http://help.keyman.com.local/developer/9.0/guides/intro
products no uc

http://help.keyman.com.local/developer/engine/android/current-version/guides/in-app/
engine no uc

### DIFF
--- a/_includes/includes/kb_doc_template.php
+++ b/_includes/includes/kb_doc_template.php
@@ -96,7 +96,7 @@
     $crumbs = explode("/",$_SERVER["REQUEST_URI"]);
     $crumbcount = count($crumbs);
     for($i=0;$i<$crumbcount;$i++){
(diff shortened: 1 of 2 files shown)


mcdurdin:
Can we Title Case the text without changing the case of the link?

darcywong00:
Ah, the links are unaffected. 

The modified lines only change the breadcrumb text. If that's intended, I can abandon this.
Title-casing instead of `ucfirst` would take some more PHP-fu.

mcdurdin:
Yes, the intent is to keep links unmodified and ucfirst the link text :grin:

PR #597, opened 2022-09-12
docs(core): add km_kbp_event function

mcdurdin:
Note: first commit just seeds the 16.0 Core docs.


developer/core/desktop/16.0/c-api/state-api.php
```diff
@@ -491,3 +491,84 @@ enum.
     </dl>
   </section>
 </article>
+
+
+
+<article id="km_kbp_process_event">
```
darcywong00, line 497:
```suggestion
<article id="km_kbp_event">
```

km_kbp_process_event is already used at l.437

darcywong00 (approved):
lgtm

PR #599, opened 2022-09-14
chore: Updates to work with Docker 🦭

### LINKED ISSUE #603: chore: handle folder vs filename conflicts for keyboard assets in .htaccess
While working on .htaccess redirects as part of keymanapp/help.keyman.com#599
I ran into complications in handling the keyboard help for
https://help.keyman.com/keyboard/khmer_angkor

The full path of the help page is:
http://help.keyman.com/keyboard/khmer_angkor/1.3/khmer_angkor.php

But the .htaccess rules can get confused by the asset folder
https://help.keyman.com/keyboard/khmer_angkor/1.3/khmer_angkor/ (note there's both khmer_angkor.php and the folder khmer_angkor/ with the latter showing permission denied)

Is it possilble to collapse all the khmer_angkor/source/help/khmer_angkor assets up to help/?

darcywong00:
Follow-on to #588 and fixes #603

Thanks to some pair-programming with @mcdurdin , this cleans up some of the redirects and has the site working with Docker.

A TODO to investigate, the keyboard links are generating a warning:

> Warning: session_start(): Cannot start session when headers already sent in /var/www/html/_includes/includes/head.php on line 3

~Also, the keyboard php extension isn't getting removed, so the redirect is loading something like:
http://localhost:8055/keyboard/georgian/1.0/georgian.php~

(edit) This is now correctly loading as
http://localhost:8055/keyboard/georgian/1.0/georgian

I copied the composer `check-links` step to `check-composer-links` so we can check links with either help.keyman.com.local or localhost:8055.


ermshiperete (approved):

.htaccess
```diff
@@ -145,22 +147,20 @@ Redirect "/bookmarklet" "/products/bookmarklet"
...
+RewriteRule "^developer/language/reference/(nul|index)(\.php)?$" "developer/language/reference/_$1.php" [END]
 
 # Remove index or index.php and redirect (and stop processing)
-RewriteCond "$1/index" -f [OR]
-RewriteCond "$1/index.php" -f
 RewriteRule "^((.+)/)?index(\.php)?$" "/$1" [R,L]
 
 # Remove .php extension and redirect
-RewriteCond "$1.php" -f
+RewriteCond "%{DOCUMENT_ROOT}/$1.php" -f
```
ermshiperete, line 156: > Also, the keyboard php extension isn't getting removed, so the redirect is loading something like:
http://localhost:8055/keyboard/georgian/1.0/georgian.php

Is this the reason? (We now expect a top-level file whereas previously it could be in a subdirectory)

.htaccess
```diff
@@ -2,8 +2,10 @@
 # NOTE: this is not very well tested...
 
 php_value include_path "/var/www/html/_includes:."
+#Not needed if local apache.conf includes this
```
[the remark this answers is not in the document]
darcywong00: It's needed with the Docker image. I left the comment cause my local development had the path set in apache.conf.

mcdurdin:
Note: I'd like to review this tomorrow hopefully. Got buried today with too many things to do...

.htaccess
```diff
@@ -48,7 +50,7 @@ RedirectMatch "/developer/engine/web/(?!faq|history|current-version|\d+\.\d+)(.*
 
 # current-version:15.0
 RewriteCond "%{DOCUMENT_ROOT}/$1/15.0" -d
-RewriteRule "^(.+)/current-version(/.*|$)" "$1/15.0$2"
+RewriteRule "^(.+)/current-version(/.*)?" "$1/15.0$2"
```
mcdurdin, line 53:
```suggestion
RewriteRule "^(.+)/current-version(/.*)?$" "$1/15.0$2"
```

We want to make sure we capture the whole suffix. I think it will anyway given `.*` is greedy by default in PCRE, but including `$` clarifies our intent which helps for future maintenance!

I think you need to copy the same pattern down for the other current-version rules.

.htaccess
mcdurdin:
```suggestion
RewriteCond "$1.php" -f
```

As @ermshiperete says, I think this should be restored to the original.

.htaccess
```diff
@@ -145,22 +147,20 @@ Redirect "/bookmarklet" "/products/bookmarklet"
...
 RewriteRule "^(.+)\.php$" "$1" [R,L]
 
 
 # Redirect folder without / to include /
-RewriteCond "$1" -d
-RewriteCond "$1.php" !-f
-RewriteCond "$1.md" !-f
+RewriteCond "%{DOCUMENT_ROOT}/$1" -d
+RewriteCond "%{DOCUMENT_ROOT}/$1.php" !-f
+RewriteCond "%{DOCUMENT_ROOT}/$1.md" !-f
```
mcdurdin, line 163:
```suggestion
RewriteCond "$1" -d
RewriteCond "$1.php" !-f
RewriteCond "$1.md" !-f
```

I don't think we want the document root requirement here.

keyboard/.htaccess
```diff
@@ -1,18 +1,14 @@
...
+RewriteRule "^gff_(mdx|muz|suq)_7(/.*)?" "/keyboard/gff_mym_7$2" [R,L]
 
 
 # Rewrite /keyboard/[name][/] to /keyboard/redir.php?q=[name]
 # only if index.php does not exist in the folder
-RewriteCond "%{DOCUMENT_ROOT}/keyboard/(?!redir.php)$1/index.php" !-f
-RewriteRule "/keyboard/([^/?]+)(/)?$" "/keyboard/redir.php?q=$1" [R,L]
-
+RewriteCond "%{DOCUMENT_ROOT}/keyboard/$1/index.php" !-f
+RewriteRule "^(?!index.php|redir.php)([^/?]+)(/)?$" "/keyboard/redir.php?q=$1" [END]
```
mcdurdin, line 9:
```suggestion
RewriteRule "^(?!index.php|redir.php)([^/?]+)(/)?$" "/keyboard/redir.php?q=$1" [R,L]
```

This was a redirect, I think it still needs to be

mcdurdin (changes requested):
I'm not 100% sure on all the redirects -- mod_rewrite is pretty hard to work with... but it's looking pretty close.

.htaccess
darcywong00: Done.

But the keyboard pages still show `.php` :disappointed:

mcdurdin: Ping me if you want to rubber-ducky on this next week?

_includes/errors/404.php
```diff
@@ -3,10 +3,11 @@
 
   head([
     'title' => "Page not found",
-    'index' => false   
+    'index' => false
   ]);
   
-  $page = $_SERVER["REQUEST_URI"];
+  // IIS uses REQUEST_URI; Apache using uri
```
ermshiperete, line 9:
```suggestion
  // IIS uses REQUEST_URI and _SERVER; Apache using uri and _REQUEST
```

I assume that's the way it is...

.htaccess
```diff
@@ -138,53 +141,71 @@ Redirect "/products/keymanweb" "/products/web"
...
 # Remove index or index.php and redirect (and stop processing)
-RewriteCond "$1/index" -f [OR]
-RewriteCond "$1/index.php" -f
-RewriteRule "^((.+)/)?index(\.php)?$" "/$1" [R,L]
+RewriteCond "{DOCUMENT_ROOT}/$1" -d
+RewriteRule "^((.+)/)?index(\.php)?$" "$1" [R,L]
 
 # Remove .php extension and redirect
-RewriteCond "$1.php" -f
+RewriteCond "{DOCUMENT_ROOT}/$1.php" -f
```
darcywong00, line 181:
```suggestion
RewriteCond "$1.php" -f
```

> As @ermshiperete says, I think this should be restored to the original.

I tried the suggestion, but the keyboard pages failed to load

darcywong00: nvm, it's working now without `{DOCUMENT_ROOT}`

darcywong00:
Currently, keyboard pages generate these Setnry [errors](https://sentry.io/organizations/keyman/issues/3592581591/?project=5983515&query=is%3Aunresolved)

> Warning: session_start(): Cannot start session when headers already sent

They seem to be caused from
https://github.com/keymanapp/help.keyman.com/blob/947929011fc35865a2bccbb5f1339bc71cdf6156/_includes/includes/head.php#L3

and go away when I remove that line?

darcywong00:
hmm, the session seems fine with no changes to head.php

.htaccess
```diff
@@ -138,53 +141,71 @@ Redirect "/products/keymanweb" "/products/web"
...
+RewriteRule "^developer/language/reference/(nul|index)(\.php)?$" "developer/language/reference/_$1.php" [END]
+
+# no pages currently reference this but will allow for future automatic re-versioning links in docs
+RewriteRule "^developer/15.0/guides/develop/creating-a-touch-keyboard-layout-for-amharic-with-keyman-developer-10(.*)?$" "developer/15.0/guides/develop/creating-a-touch-keyboard-layout-for-amharic.php" [END]
 
 # Remove index or index.php and redirect (and stop processing)
-RewriteCond "$1/index" -f [OR]
-RewriteCond "$1/index.php" -f
-RewriteRule "^((.+)/)?index(\.php)?$" "/$1" [R,L]
+RewriteCond "{DOCUMENT_ROOT}/$1" -d
```
mcdurdin, line 177: This rule doesn't look correct? Does it need `{DOCUMENT_ROOT}`?

mcdurdin (approved):

PR #600, opened 2022-09-14
feat:  docs for wordbreaker extension API

jahorton:
This PR documents the new features added by https://github.com/keymanapp/keyman/pull/7279, which facilitates wordbreaker extension and customization.


developer/15.0/guides/lexical-models/advanced/unicode-breaker-extension.md
```diff
@@ -0,0 +1,250 @@
...
+
+As written in the spec, WB6 reads **AHLetter x (MidLetter | MidNumLetQ) AHLetter**.
+
+To break that down:
+- (implicit) **Any**
+- **AHLetter**: `ALetter` or `Hebrew_Letter`
+- **x** - "do not break"
+- (either) **MidLetter** or **MidNumLetQ**: one of `MidNumLet` or `Single_Quote`
+- **AHLetter**: `ALetter` or `Hebrew_Letter`
+    - These expansions are defined at https://unicode.org/reports/tr29/#WB_Rule_Macros.
```
mcdurdin, line 167:
```suggestion

The expansions `MidNumLetQ` and `AHLetter` are defined at https://unicode.org/reports/tr29/#WB_Rule_Macros.
```

developer/15.0/guides/lexical-models/advanced/unicode-breaker-extension.md
```diff
@@ -0,0 +1,250 @@
...
+
+If written as a custom rule, it takes the following form:
```
mcdurdin, line 169:
```suggestion
If written as a custom rule, it takes the following form, using the function `context.propertyMatch`, which takes 4 parameters, matching two characters before and two characters after a potential boundary:
```

developer/15.0/guides/lexical-models/advanced/unicode-breaker-extension.md
```diff
@@ -0,0 +1,250 @@
...
+
+```typescript
+{
+  match: (context) => {
+    return context.propertyMatch(null, // no requirements:  implicit "Any".
```
mcdurdin, line 174:
```suggestion
    return context.propertyMatch(null, // match any character:  implicit "Any".
```

developer/15.0/guides/lexical-models/advanced/unicode-breaker-extension.md
```diff
@@ -0,0 +1,250 @@
...
+Almost all of the specification's rules are of this form:  up to two characters
+before a potential boundary may be considered alongside up to two characters after
+the potential boundary.
+
+#### Word-breaking property names
+The names used in each array must be defined in one of the
+following places:
+* https://unicode.org/reports/tr29/#Table_Word_Break_Property_Values (case-insensitive)
+* `customProperties` - your <a href="#define">declaration of any custom property types</a>
+* This list: `["Other", "sot", "eot"]`
```
mcdurdin, line 193:
```suggestion
* One of the special property types `"Other"`, `"sot"`, or `"eot"`:
```

developer/15.0/guides/lexical-models/advanced/unicode-breaker-extension.md
```diff
@@ -0,0 +1,250 @@
...
+    * `eot`:  "end of text" - a marker indicating the end of the string being word-broken
+
+#### Applying the example
+As an example, when determining whether or not to break the English word `don't` when
+applying WB6 as written above, this is what happens near the apostrophe:
+
+```typescript
+{
+  match: (context) => {
+    return context.propertyMatch(null, /* automatic match */                // "o" - ALetter
```
mcdurdin, line 205:
```suggestion
    return context.propertyMatch(null, /* match any character */            // "o" - ALetter
```

developer/15.0/guides/lexical-models/advanced/unicode-breaker-extension.md
```diff
@@ -0,0 +1,250 @@
...
+                                // x
+                                ["MidLetter", "MidNumLet", "Single_Quote"], // "t" - Single_Quote
+                                ["ALetter", "Hebrew_Letter"]);              //     - eot
+  },
+  breakIfMatch: false // do not break
+}
+```
+
+The rule, however, does not apply between the `'` and the `t`, so without the spec's WB7 rule,
+a wordbreak could occur here once WB999 is reached.
```
mcdurdin, line 232: I don't really understand what you are saying here?

mcdurdin (commented):
This is well written and really helpful. I've added a bunch of suggestions and formatting edits and asked for clarity in a couple of places.

developer/15.0/guides/lexical-models/advanced/unicode-breaker-extension.md
```diff
@@ -0,0 +1,250 @@
...
+};
+
+export default source;
+```
+
+This example's customization is designed to accomplish two goals:
+
+1. Unicode's wordbreaker does not map base Khmer consonants to any of the
+  <a href="https://unicode.org/reports/tr29/#Table_Word_Break_Property_Values">
+  relevant wordbreaking character properties</a>, causing it to be treated as `"Other"`.
```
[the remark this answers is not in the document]
jahorton: Late-day brain just couldn't remember the Markdown for doing links.  :+1:, will fix other cases.

developer/15.0/guides/lexical-models/advanced/word-breaker.php
```diff
@@ -34,6 +34,9 @@ function</dfn> is responsible for. It is a little bit of code that looks at some
        you can <a href="#join">choose which strings join words together</a>. </li>
   <li> If the default word breaker creates <strong>not enough splits</strong>,
        you must <a href="#custom">create your own word breaker function</a>. </li>
+  <li> Alternatively, you may choose to <a href="./unicode-breaker-extension">
```
darcywong00, line 37: Can maybe also add a cross-reference in "See also" l. 237

developer/15.0/guides/lexical-models/advanced/unicode-breaker-extension.md
jahorton: I've reworked that area a bit; hopefully it's clearer now?

developer/15.0/guides/lexical-models/advanced/unicode-breaker-extension.md
```diff
@@ -0,0 +1,250 @@
...
+
+2. By default, the wordbreaker will automatically insert word boundaries before
+  and after a hyphen.  To allow mid-word hyphens, we need to remap them to
+  a different character property.
+
+  While the specification itself mentions that we could
+  just map hyphens to `"MidLetter"`, this example opts to define
+  a custom-tailored property that ensures only hyphens are affected in order to add
+  a special, end-of-context rule that may be useful when typing - the rule above
+  referencing `"eot"`.
```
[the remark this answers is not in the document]
jahorton: This is covered more explicitly in one of the newly-written sections.

developer/15.0/guides/lexical-models/advanced/unicode-breaker-extension.md
```diff
@@ -0,0 +1,563 @@
...
+    - It does not include any of the `MidLetter` characters.
+
+Matching the rule against the end of text suggests that what follows the `Hyphen` character
+is the point of text insertion.  For this example, assuming that a user has just typed
+`full-`, there will be no word-break on the hyphen until either more input is received or
+the user changes the site of text entry.
+
+**Important note**: you must declare any custom properties within the `customProperties` array.
+If any are missing, the missing custom properties will fail to match against any word-breaking rule.
+Make sure you don't misspell it anywhere in your customization code!
```
jahorton, line 436: Revisiting this aspect and writing this note reminded me of something I've felt is "a bit off."

See https://github.com/keymanapp/keyman/pull/7279#discussion_r971493917 - I think we should be able to infer the properties, rather than require a declaration.  That said.

> Make sure you don't misspell it anywhere in your customization code!

Yeah, that part will still need to stick around.

developer/15.0/guides/lexical-models/advanced/unicode-breaker-extension.md
```diff
@@ -0,0 +1,563 @@
...
+```
+
+Let's walk through what this simplification is trying to achieve:
+
+1. Hyphens are mapped to their own distinct word-breaking property.
+2. The custom rule prevents wordbreaking between a letter and a hyphen at the end of text.
+    - It does not include any of the `MidLetter` characters.
+
+Matching the rule against the end of text suggests that what follows the `Hyphen` character
+is the point of text insertion.  For this example, assuming that a user has just typed
```
jahorton, line 430: I'd need to add unit tests to be sure, but I believe this isn't just a suggestion - it's a practical reality!

Reference:  https://github.com/keymanapp/keyman/blob/4f4ddb61078929e9d26db2991b502d8db1073702/common/models/templates/src/tokenization.ts#L26-L34

Also of note:  this could be used as a "default extension" to achieve what this part of that function is aiming to do: https://github.com/keymanapp/keyman/blob/4f4ddb61078929e9d26db2991b502d8db1073702/common/models/templates/src/tokenization.ts#L52-L72

developer/15.0/guides/lexical-models/advanced/unicode-breaker-extension.md
jahorton: Started implementing the enhancement, then realized some limitations of the approach that posed significant issues.  I'll probably leave things as they are instead, and thus leave this note as-is.

developer/15.0/guides/lexical-models/advanced/unicode-breaker-extension.md
jahorton: Related unit tests: https://github.com/keymanapp/keyman/pull/7279/commits/26fb6da3abdd1df7c6a9f36ed9aaa0797468da9d

Note:  no changes were needed to the tokenization method; I was right!

developer/15.0/guides/lexical-models/advanced/unicode-breaker-extension.md
```diff
@@ -228,23 +240,324 @@ word-breaker not to word-break if the rule matches at that location.
 }
 ```
 
-The rule, however, does not apply between the `'` and the `t`, so without the spec's WB7 rule,
-a wordbreak could occur here once WB999 is reached.
+The same rule does not apply between the `'` and the `t`, so it does not apply at this position in
+the text.This is why the wordbreaking spec includes rule WB7 in addition to WB6 - some scenarios
```
mcdurdin, line 244:
```suggestion
the text. This is why the wordbreaking spec includes rule WB7 in addition to WB6 - some scenarios
```

developer/15.0/guides/lexical-models/advanced/unicode-breaker-extension.md
```diff
@@ -215,6 +225,8 @@ applying WB6 as written above, this is what happens near the apostrophe:
 The rule applies at the position between the `n` and the `'` of `don't`, telling the
 word-breaker not to word-break if the rule matches at that location.
 
+#### An unsuccessful rule match
+
```
mcdurdin, line 229:
```suggestion

Given the input `n't|` (where `|` is the caret), we can see that the rule will not match:
```

developer/15.0/guides/lexical-models/advanced/unicode-breaker-extension.md
```diff
@@ -228,23 +240,324 @@ word-breaker not to word-break if the rule matches at that location.
...
+</p>
+</aside>
 
 Defining custom rules is a powerful tool, but it is detailed work and may be somewhat tedious
-to get right.  Feel free to <a href="https://community.software.sil.org/c/keyman/19">ask for
-help at our Community Site </a> for assistance.
+to get right.  Feel free to [ask for help at our Community Site](
+https://community.software.sil.org/c/keyman/19) for assistance.
+
+## <a name="map" id="map"></a> Character property remapping
```
mcdurdin, line 263: Note that Markdown gives headings an id already, so this could be linked to with `#character-property-remapping`, making the `<a...>` unnecessary.

developer/15.0/guides/lexical-models/advanced/unicode-breaker-extension.md
```diff
@@ -228,23 +240,324 @@ word-breaker not to word-break if the rule matches at that location.
...
+},
+```
+
+If you search the [property definition text file](
+https://www.unicode.org/Public/UCD/latest/ucd/auxiliary/WordBreakProperty.txt)
+for `1780` or `17b3`, you will find neither.  These correspond to many letters
+from the Khmer character set - notably, the 'base' characters used in Khmer's
+grapheme clusters.  The other Khmer characters tend to attach at various positions
+around these base characters.  The majority language for the script - Khmer -
+does not follow conventionalword-breaking rules; most notably, they do not add
```
mcdurdin, line 319:
```suggestion
does not follow conventional word-breaking rules; most notably, they do not add
```

mcdurdin (approved):
Changes LGTM. Minor formatting and typos only.

developer/15.0/guides/lexical-models/advanced/unicode-breaker-extension.md
```diff
@@ -0,0 +1,564 @@
...
+```
+
+Note that we have left out some of the rule extensions that would help cover certain
+less-frequently encountered cases.  These may matter to some parts of a
+language community, especially for models targeting a majority language.
+Computer programmers in particular tend to care about the un-extended
+rules (WB9, 10, 13a, and 13b) more than most.
+
+Remember, remapping characters to a new word-breaking property prevents any default
+rule from handling them unless you add custom rules to re-includes them as their
```
jahorton, line 559:
```suggestion
rule from handling them unless you add custom rules to re-include them as their
```

developer/15.0/guides/lexical-models/advanced/word-breaker.php
```diff
@@ -230,6 +233,9 @@ class RegExpDerivedSpan implements Span {
 <p>
   <a href="https://github.com/keymanapp/keyman/blob/4211b468949860b8fb4a4707710472ab9e33c581/common/lexical-model-types/index.d.ts#L286-L323">
 The TypeScript definition of <code>WordBreakingFunction</code> and <code>Span</code></a>
+  <a href="./unicode-breaker-extension.md">Extension and customization of the Unicode word-breaker</a>
```
jahorton, line 236:
```suggestion
  <a href="./unicode-breaker-extension">Extension and customization of the Unicode word-breaker</a>
```

Was getting a "broken link" report from this!

mcdurdin: Yes, policy on the websites is to exclude both `.md` and `.php` extensions -- bare filenames only. We have some legacy `.php` links but `.md` will intentionally always fail :grin:, so that links are independent of underlying format. (For legacy reasons, links to `.../foo.php` will forward correctly to `.../foo` when we convert `foo.php` to `foo.md`, and then our .htaccess will pass that to the markdown renderer transparently!).

eddieantonio (approved):
LGTM!

jahorton:
OK, so one question I do have before merging - should I actually merge this as-is?  This is being applied to 15.0's docs, but 15.0 won't have the feature.

mcdurdin:
> should I actually merge this as-is? This is being applied to 15.0's docs, but 15.0 won't have the feature.

Good catch. You are right, we'll need to copy the `/developer/15.0` folder to `/developer/16.0` (and fix any 15.0 references in the 16.0 copy), and then shift this PR to target those 16.0 files.

darcywong00:
> we'll need to copy the `/developer/15.0` folder to `/developer/16.0`...

That will also take care of keymanapp/keyman#7282

mcdurdin:
Note: I moved the content to 16.0 docs

PR #601, opened 2022-09-15
chore: Update ci.yml from Windows to Linux 🦭

darcywong00:
Follow-on to #599 

This updates the GHA from Windows to ubuntu-20.04 ([ubuntu-latest](https://github.com/marketplace/actions/setup-php-action#github-hosted-runners) uses PHP 8.1 and doesn't have 7.4)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ jobs:
   build-test:
     runs-on: ubuntu-20.04
     env:
-      KEYMANHOSTS_TIER: TIER_TEST
+      KEYMANHOSTS_TIER: test
 
     steps:
     - name: Checkout
@@ -28,7 +28,7 @@ jobs:
     - name:  Install dependencies
       shell: bash
       run: |
-        echo "$TIER_TEST" > tier.txt
+        echo "test" > tier.txt
         composer install --no-progress
       env:
         fail-fast: true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,7 @@ jobs:
       shell: bash
       run: |
         ( set +e; set +o pipefail; npx broken-link-checker http://localhost:8055 --ordered --recursive --host-requests 50 -e --filter-level 3; exit ${PIPESTATUS[0]} )
-
+        docker container logs `docker container ls -ls -q`
       # ( set +e; set +o pipefail; npx broken-link-checker http://localhost:8055 --ordered --recursive --host-requests 50 -e --filter-level 3 | grep -E "BROKEN|Getting links from" | grep -B 1 "BROKEN"; exit ${PIPESTATUS[0]} )
 
     - name: Check PHP errors
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,8 @@ jobs:
     - name: Check broken links
       shell: bash
       run: |
-        ( set +e; set +o pipefail; npx broken-link-checker http://localhost:8055 --ordered --recursive --host-requests 50 -e --filter-level 3; exit ${PIPESTATUS[0]} )
+        # ( set +e; set +o pipefail; npx broken-link-checker http://localhost:8055 --ordered --recursive --host-requests 50 -e --filter-level 3; exit ${PIPESTATUS[0]} )
+        curl http://localhost:8055
         docker container logs `docker container ls -ls -q`
       # ( set +e; set +o pipefail; npx broken-link-checker http://localhost:8055 --ordered --recursive --host-requests 50 -e --filter-level 3 | grep -E "BROKEN|Getting links from" | grep -B 1 "BROKEN"; exit ${PIPESTATUS[0]} )
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,13 +65,13 @@ jobs:
     # If any of the tests fail, let's grab a bit more detail on the environment
     #
     - name: Report errors
-      if: ${{ failure() }}
+      #if: ${{ failure() }}
       shell: bash
       run: |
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -5,7 +5,7 @@ on: [pull_request]
 jobs:
 
   build-test:
-    runs-on: windows-latest
+    runs-on: ubuntu-20.04
```
darcywong00, line 8: ubuntu-latest uses PHP 8.1 instead of 7.4

.github/workflows/ci.yml
```diff
@@ -22,71 +22,25 @@ jobs:
...
     #
     - name:  Install dependencies
-      shell: cmd
+      shell: bash
       run: |
-        echo TIER_TEST > tier.txt
+        echo "$TIER_TEST" > tier.txt
         composer install --no-progress
-        npm install
+      # npm install
```
darcywong00, line 33: TODO: I don't see where `package.json` is

darcywong00:
ugh, running the link checker through Docker seems really slow (6+ hours before GH timed out)

.github/workflows/ci.yml
```diff
@@ -22,84 +22,39 @@ jobs:
...
     #
     - name:  Install dependencies
-      shell: cmd
+      shell: bash
       run: |
-        echo TIER_TEST > tier.txt
+        echo "$TIER_TEST" > tier.txt
         composer install --no-progress
-        npm install
+      env:
```
darcywong00, line 33: the previous DOS cmd ran composer.bat so it needed the `npm install`.
Bash no longer needs this

.htaccess
```diff
@@ -150,19 +150,22 @@ Redirect "/bookmarklet" "/products/bookmarklet"
 # Rewrite nul -> _nul, index -> _index
 RewriteRule "^developer/language/reference/(nul|index)(\.php)?$" "developer/language/reference/_$1.php" [END]
 
+# no pages currently reference this but will allow for future automatic re-versioning links in docs
+RewriteRule "^developer/15.0/guides/develop/creating-a-touch-keyboard-layout-for-amharic-with-keyman-developer-10(.*)?$" "developer/15.0/guides/develop/creating-a-touch-keyboard-layout-for-amharic.php" [END]
+
```
darcywong00, line 155: From https://github.com/keymanapp/help.keyman.com/blob/947929011fc35865a2bccbb5f1339bc71cdf6156/developer/15.0/guides/develop/web.config#L6-L9

I had to include in the root `.htaccess` because `/developer/15.0/guides/develop/.htaccess` wasn't running inherited rewrite rules

Dockerfile
```diff
@@ -1,7 +1,5 @@
 FROM php:7.4-apache
 RUN a2enmod rewrite
 
-RUN sed -ri -e 's!DirectoryIndex index.php index.html!DirectoryIndex index.md index.php index.html!g' -e 's!</FilesMatch>!</FilesMatch>\n\n<FilesMatch \\.md$>\n\tSetHandler text/html\n</FilesMatch>!g' /etc/apache2/conf-available/docker-php.conf
-
-# ENV APACHE_LOGLEVEL trace3
-# RUN sed -ri -e 's!LogLevel warn!LogLevel warn rewrite:${APACHE_LOGLEVEL}}!g' /etc/apache2/apache2.conf
+RUN sed -ri -e "s!DirectoryIndex index.php index.html!DirectoryIndex index.md index.php index.html!g" -e "s!</FilesMatch>!</FilesMatch>\n\n<FilesMatch \\.md$>\n\tSetHandler text/html\n</FilesMatch>!g" "/etc/apache2/conf-available/docker-php.conf" \
+ && sed -ri -e "s!</VirtualHost>!\n        <Directory /var/www/html>                \n                Options +Includes +FollowSymLinks -MultiViews\n                AllowOverride All\n        </Directory>\n\n        php_value include_path \"/var/www/html/_includes:.\"\n</VirtualHost>!g" "/etc/apache2/sites-available/000-default.conf"
```
darcywong00, line 5: Key changes: `-MultiViews` and setting the include_path

.github/workflows/ci.yml
darcywong00: On Windows, cmd ran composer.bat so `npm install` needed to be called.
On Linux, we don't need this anymore

_common/KeymanHosts.php
```diff
@@ -90,6 +90,8 @@ function __construct() {
         $site_suffix = '.local';
         $site_protocol = 'https://';
         break;
+      default:
+        die("tier is '$this->tier' which is invalid\n");
```
darcywong00, line 94: An invalid tier would mean `$site_suffix` and `$site_protocol` never get set, leading to a really slow link checker.

.htaccess
```diff
@@ -172,15 +175,11 @@ RewriteRule "^(.+[^/])$" "$1/" [R,L]
 RewriteCond "%{DOCUMENT_ROOT}/$1.md" -f
 RewriteRule "^(.+)$" "/_includes/md/mdhost.php?file=$1.md" [END]
 
-RewriteCond "%{DOCUMENT_ROOT}/$1.md" -f
-RewriteRule "^(.+)\.md$" "/_includes/md/mdhost.php?file=$1.md" [END]
-
```
darcywong00, line 177: I thought I needed this. The original web.config doesn't have this.

.github/workflows/ci.yml
mcdurdin: Note: not so much that it needed `npm install` but rather that `npm install` was never run because `composer` was a batch file, so the batch script drops out after running it (it would need to have been `call composer install --no-progress` for the next step to run. Perils of batch files!

mcdurdin (approved):
LGTM. Next step: setting up new staging sites with @tim-eves help?

PR #602, opened 2022-09-15
fix: Include requested URI on 404 error 🦭

darcywong00:
Follow-on to #588

On Apache, this passes the requested URI for 404 errors. (IIS still using REQUEST_URI)

afaict, Apache redirects ErrorDocument so the original code:

```
$page = $_SERVER["REQUEST_URI"]
```
was displaying the path to the 404.php page.

atm I don't know how to prevent the addressbar from updating
e.g.
http://localhost.8055/bogus results in
http://localhost:8055/_includes/errors/404?uri=/bogus


mcdurdin (approved):
LGTM. I am okay with the 404 redirect instead of rewrite ... it's not ideal, but it's for a 404 page so it's not the end of the world.

(It's annoying for advanced users who'd just edit the url that they presumably mis-typed, but now have to get rid of the 404 page... Looked online but didn't find any good answers on how to avoid this.)

PR #604, opened 2022-09-27
chore: Add builder scripts for Developers 🦭

darcywong00:
Follow-on to #599 and joining the [build-utils](https://github.com/keymanapp/keyman/blob/master/resources/build/build-utils.md) movement

~The first commit is seeding the builder scripts from the main repo.
I then removed the VERSION and Sentry bits.~
Updated to use builder.inc.sh from keymanapp/keyman#7365

Rather than having to memorize all the Docker incantations:

`./build.sh configure` - Installs dependencies
`./build.sh build` - Downloads docker image and starts the container. Once this is called, the site is accessible at http://localhost:8055
`./build.sh test` - Runs the link checker. Do we need to include lint?

Other  possible TODOs
- [x] Verify this works on Windows host
- [ ] Verify this works on macOS host (coreutils dependency?)
https://github.com/keymanapp/keyman/blob/497a8613db42fee2f812e4a5d48b11651bbd9eae/resources/build/build-utils.sh#L46
* build.sh clean - I haven't sorted how to clean up Docker images/containers.


darcywong00:
At what point should tier.txt be created?

resources/shellHelperFunctions.sh
```diff
@@ -0,0 +1,292 @@
+#!/usr/bin/env bash
```
mcdurdin, line 1: Let's not include this file. We shouldn't need it for this environment.

mcdurdin (commented):
Conceptually I really like the idea.

One concern: we'll have to keep build-utils.sh in sync between repos. I am not sure how we best handle this, but read below for some potential mitigations.

There are also some assumptions in build-utils.sh that we should probably revisit before we go too far with using it cross-repo, e.g.:

* `$KEYMAN_ROOT` variable name is misleading and should probably be `$REPO_ROOT`
* We should remove any code specific to the keyman repo, e.g. `findTier`, `exportEnvironmentDefinitionScript` (and thus don't need that entry in .gitignore), `set_keyman_standard_build_path`, `printXCodeBuildScriptLogs`, `run_xcodebuild` ...

In fact, it might be better to split the builder_ script stuff into /resources/builder.inc.sh, on the Keyman repo side, and source include it from build-utils.sh (`. "$KEYMAN_ROOT/resources/builder.inc.sh`), and aim to make builder.inc.sh a cross-repo compatible script, including defining `$REPO_ROOT`.

Once we do that, I would also rename `findRepositoryRoot` to `_builder_findRepoRoot`, and rename `die` to `builder_die` (and see if we can implement color in it also).

Don't set `SHLVL` in builder.inc.sh -- it should be as pure as possible in its functionality.

At the end of the day, we'll `. "$(dirname "$THIS_SCRIPT")/../resources/builder.inc.sh"` from scripts in other repos, and one day do the same in scripts in Keyman repo, and avoid the `build/` folder (which also has problems because it's a standard build output path name...)

build.sh
```diff
@@ -0,0 +1,70 @@
...
+  fi
+
+  builder_finish_action success build
+fi
+
+if builder_start_action test; then
+  # TODO Lint?
+
+  # Check for broken links
+  composer check-docker-links
```
darcywong00, line 69: Do we want this to be consistent with the GitHub action?

https://github.com/keymanapp/help.keyman.com/blob/191fe6144d1cfbf2557761c277c420e6c4030cb3/.github/workflows/ci.yml#L94-L107

build.sh
```diff
@@ -0,0 +1,70 @@
...
+  "Setup help.keyman.com site to run via Docker." \
+  configure \
+  clean \
+  build \
+  test \
+
+builder_parse "$@"
+
+# START - Script parameter configuration
+REPORT_STYLE=local  # Default setting.
```
darcywong00, line 31: Maybe this can be removed?

build.sh
```diff
@@ -0,0 +1,70 @@
...
+#
+set -eu
+
+## START STANDARD BUILD SCRIPT INCLUDE
+# adjust relative paths as necessary
+THIS_SCRIPT="$(greadlink -f "${BASH_SOURCE[0]}" 2>/dev/null || readlink -f "${BASH_SOURCE[0]}")"
+. "$(dirname "$THIS_SCRIPT")/resources/build/builder.inc.sh"
+## END STANDARD BUILD SCRIPT INCLUDE
+
+. "$REPO_ROOT/resources/shellHelperFunctions.sh"
```
darcywong00, line 13: Needed for get_builder_OS (l.54)

build.sh
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env bash
+#
+# Setup help.keyman.com site to run via Docker.
+#
+set -eu
+
+## START STANDARD BUILD SCRIPT INCLUDE
+# adjust relative paths as necessary
+THIS_SCRIPT="$(greadlink -f "${BASH_SOURCE[0]}" 2>/dev/null || readlink -f "${BASH_SOURCE[0]}")"
+. "$(dirname "$THIS_SCRIPT")/resources/build/builder.inc.sh"
```
darcywong00, line 10: Note: using builder.inc.sh instead of "standard" build-utils.sh

build.sh
```diff
@@ -0,0 +1,70 @@
...
+# START - Script parameter configuration
+REPORT_STYLE=local  # Default setting.
+
+# END - Script parameter configuration
+
+# This script runs from its own folder
+cd "$REPO_ROOT"
+
+if builder_start_action configure; then
+  #verify_npm_setup
```
darcywong00, line 39: Is this also needed? (from cy.yml)
https://github.com/keymanapp/help.keyman.com/blob/191fe6144d1cfbf2557761c277c420e6c4030cb3/.github/workflows/ci.yml#L87-L88

resources/shellHelperFunctions.sh
darcywong00: Needed for get_builder_OS() unless that also gets moved to builder.inc.sh
`os_id` becomes readonly?

build.sh
mcdurdin: * We wouldn't want tier.txt. That's only for CI testing.
* `composer install` is good (don't include `--no-progress` flag I think)
* Delete `#verify_npm_setup`

.gitignore
```diff
@@ -5,3 +5,7 @@ cdn/deploy
 vendor/
 .vscode/
 node_modules/
+
+# Auto-generated resource script
+resources/environment.sh
```
mcdurdin, line 10: This shouldn't be needed.

Readme.md
```diff
@@ -1,114 +1,27 @@
...
-ServerName 127.0.0.1
-```
-
-Setup Local hosts file
-```bash
-sudo vim /etc/hosts
-# Add:
-127.0.0.1 help.keyman.com.local
-```
+On Windows, also install [WSL2](https://ubuntu.com/tutorials/install-ubuntu-on-wsl2-on-windows-10#1-overview)
```
mcdurdin, line 12: It's possible to run Docker on hyper-v as an alternative to WSL2

build.sh
mcdurdin: Note, builder.inc.sh should be in /resources/, not /resources/build/

build.sh
mcdurdin: Let's avoid pulling this in. shellHelperFunctions is full of Keyman-specific stuff. It's a one liner to detect Windows:

```bash
if [[ $OSTYPE =~ msys|cygwin ]]; then
  ... windows code
fi
```

build.sh
mcdurdin: Yes, remove it; it comes straight out of common\web\keyboard-processor\build.sh in the Keyman repo.

build.sh
```diff
@@ -0,0 +1,70 @@
...
+  clean \
+  build \
+  test \
+
+builder_parse "$@"
+
+# START - Script parameter configuration
+REPORT_STYLE=local  # Default setting.
+
+# END - Script parameter configuration
```
mcdurdin, line 33:
```suggestion
```

build.sh
mcdurdin, line 39:
```suggestion
```

build.sh
```diff
@@ -0,0 +1,70 @@
...
+
+if builder_start_action configure; then
+  #verify_npm_setup
+  composer install
+  
+  builder_finish_action success configure
+fi
+
+if builder_start_action clean; then
+  #npm run clean
```
mcdurdin, line 46:
```suggestion
```

build.sh
mcdurdin: Probably a good idea. Then we can just call build.sh from ci.yml

resources/shellHelperFunctions.sh
mcdurdin: If we were to do this, we'd call it `builder_os` variable. We'd keep `os_id` in shellHelperFunctions.sh and slowly fixup existing scripts to use `builder_os`.

mcdurdin (changes requested):
A few changes requested here. Certainly moving in the right direction, thanks!

build.sh
```diff
@@ -0,0 +1,69 @@
...
+  configure \
+  clean \
+  build \
+  test \
+
+builder_parse "$@"
+
+# START - Script parameter configuration
+
+# END - Script parameter configuration
```
mcdurdin, line 30:
```suggestion
```

build.sh
```diff
@@ -0,0 +1,69 @@
...
+# adjust relative paths as necessary
+THIS_SCRIPT="$(greadlink -f "${BASH_SOURCE[0]}" 2>/dev/null || readlink -f "${BASH_SOURCE[0]}")"
+. "$(dirname "$THIS_SCRIPT")/resources/builder.inc.sh"
+## END STANDARD BUILD SCRIPT INCLUDE
+
+################################ Main script ################################
+
+# Ensures color var use in `builder_describe`'s argument respects the specified
+# --color/--no-color option.
+builder_check_color "$@"
```
mcdurdin, line 17:
```suggestion
```

Not required here -- this is only needed if we use colour in the builder_describe call.

TODO: a future tweak could make this entirely internal within the builder.inc.sh script, so that we don't have to jump through this hoop for build.sh scripts: just call it at the same time as we do _builder_init (and then rename to _builder_check_color_param)

build.sh
```diff
@@ -0,0 +1,69 @@
...
+  composer install
+  
+  builder_finish_action success configure
+fi
+
+if builder_start_action clean; then
+  # Stop and cleanup all docker containers/images
+  docker kill $(docker ps -q)
+  docker rm $(docker ps -a -q)
+  docker rmi $(docker images -q)
```
mcdurdin, line 45: Isn't this dangerous? Won't it kill *all* docker containers and images, even unrelated ones? If so, that could be pretty disastrous!

mcdurdin (changes requested):
Some small tweaks, but one big thing: I'm pretty uncomfortable with the `clean` action right now ... am I reading it right?

build.sh
darcywong00: Yeah, the clean is pretty excessive. I'll investigate if I can limit it to `keyman-websites` images.

darcywong00: k, I've updated clean to only remove `keyman-website` images. Though this may impact the local keyman.com site if it also uses the same image

build.sh
```diff
@@ -39,31 +31,43 @@ if builder_start_action configure; then
...
+  docker build -t help-keyman-website .
 
   if [[ $OSTYPE =~ mysys|cygwin ]]; then
     # Windows needs leading slashes for path
-    docker run -d -p 8055:80 -v //$(pwd):/var/www/html/ -e S_KEYMAN_COM=localhost:8054 keyman-websites
+    docker run -d -p 8055:80 -v //$(pwd):/var/www/html/ -e S_KEYMAN_COM=localhost:8054 help-keyman-website
   else
-    docker run -d -p 8055:80 -v $(pwd):/var/www/html/ -e S_KEYMAN_COM=localhost:8054 keyman-websites
+    docker run -d -p 8055:80 -v $(pwd):/var/www/html/ -e S_KEYMAN_COM=localhost:8054 help-keyman-website
   fi
```
mcdurdin, line 54: This should probably be under a `./build.sh start` action, and we could have a corresponding `stop` action (from `clean`, although probably still do the `docker container stop` in the `clean` action)?

Question if `stop` should `docker container rm` as well?

mcdurdin (changes requested):
Looking very close. Just minor action suggestions to round this out.

build.sh
darcywong00: ok. I've added start and stop actions.

I only have the clean action rm containers and images.

build.sh
```diff
@@ -12,11 +12,30 @@ THIS_SCRIPT="$(greadlink -f "${BASH_SOURCE[0]}" 2>/dev/null || readlink -f "${BA
...
+
+function _stop_docker_container() {
+  HELP_CONTAINER=$(_get_docker_container_id)
+  if [ ! -z "$HELP_CONTAINER" ]; then
+    docker container stop $HELP_CONTAINER
+  else
+    echo "No Docker container to stop"
+  fi
+}
+
```
mcdurdin, line 31: ❤️ These are great! We can generalize when we start using across multiple sites.

I see a need for us to figure out a good way to share between repos -- this and other scripts. I think build.sh will need to do this one day in the configure step? Having some ideas, won't block this PR, but should solve both this and the common PHP files.

build.sh
```diff
@@ -46,28 +74,36 @@ if builder_start_action build; then
...
+  if [ ! -z $(_get_docker_container_id) ]; then
+    if [[ $OSTYPE =~ mysys|cygwin ]]; then
+      # Windows needs leading slashes for path
+      docker run -d -p 8055:80 -v //$(pwd):/var/www/html/ -e S_KEYMAN_COM=localhost:8054 help-keyman-website
+    else
+      docker run -d -p 8055:80 -v $(pwd):/var/www/html/ -e S_KEYMAN_COM=localhost:8054 help-keyman-website
+    fi
   else
-    docker run -d -p 8055:80 -v $(pwd):/var/www/html/ -e S_KEYMAN_COM=localhost:8054 help-keyman-website
+    echo "${COLOR_RED}ERROR: Docker container doesn't exist. Run ./build.sh build first"
```
mcdurdin, line 90:
```suggestion
    echo "${COLOR_RED}ERROR: Docker container doesn't exist. Run ./build.sh build first${COLOR_RESET}"
```

:+1:

build.sh
```diff
@@ -46,28 +74,36 @@ if builder_start_action build; then
...
-  builder_finish_action success build
+  builder_finish_action success start
 fi
 
-if builder_start_action test; then
+if builder_start_action stop; then
+  # Stop the Docker container
+  _stop_docker_container
+  builder_finish_action success stop
+fi
```
mcdurdin, line 101: Suggest putting this before `build` action, so you can do:

`./build.sh stop build start`

and it will run the actions in the order the user expects.

Conversely, I can't see `./build.sh start stop` as a sensible thing really...

Note: I know this is a little bit unintuitive at present -- build script actions are always run in a specific order, regardless of the order in which they are specified!

mcdurdin (changes requested):
Super close now :grin:

Readme.md
```diff
@@ -13,17 +13,26 @@ On Windows, Docker will also need either:
...
+1. Run `./build.sh start`.
 
-1. Run `./build.sh build`
+After this, you can access the help.keyman site at http://localhost:8055
 
-## Running tests
+#### Stop the Docker container
+1. Run `./build.sh stop`
 
+#### Clean the Docker container and image
```
mcdurdin, line 33:
```suggestion
#### Remove the Docker container and image
```

PR #611, opened 2022-10-11
Update index.php

SabineSIL:
no link to previous page available


mcdurdin (changes requested):
Probably better to add the previous page link :grin:

mcdurdin:
Replaced with #615

PR #613, opened 2022-10-11
Update step-7.php

SabineSIL:
Line 23: There is no entry F5 in the menu in Keyman-Developer -> Keyboard -> Test Keyboard.   "Test Keyboard F5"  would be nice

### DIFF
--- a/developer/15.0/guides/develop/tutorial/step-7.php
+++ b/developer/15.0/guides/develop/tutorial/step-7.php
@@ -19,8 +19,8 @@ occurred.</p>
 
 <h2>Compiling the Keyboard</h2>
 
-<p>After compiling the keyboard successfully, we can start testing it. Choose <span class="guimenu">Keyboard</span>, <span class=
-"guimenuitem">Start Test</span> or press <kbd>F5</kbd> to begin testing. The Keyboard test window will appear.</p> 
+<p>After compiling the keyboard successfully, we can start testing it. Choose <span class="guimenu">Debug</span>, <span class=
+"guimenuitem">Start Debugging</span> or press <kbd>F5</kbd> to begin testing. The Keyboard debug window will appear.</p> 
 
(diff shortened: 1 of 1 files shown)


mcdurdin:
There's no change made here. F5 does work, even if it isn't shown.

SabineSIL:
Inside the menu of Keyman-Developer -> Keyboard  we have the line "Test Keyboard"  It would be nicer for the user to have  "**Test Keyboard     F5**" instead. ( As it is done for "Compile Keyboard    F7" or "Install...  Ctrl+F9" )

mcdurdin:
Ah, I see. The <kbd>F5</kbd> shortcut is used with `Start Debugging` in the `Debug` menu instead.

developer/15.0/guides/develop/tutorial/step-7.php
```diff
@@ -20,7 +20,7 @@
 <h2>Compiling the Keyboard</h2>
 
 <p>After compiling the keyboard successfully, we can start testing it. Choose <span class="guimenu">Keyboard</span>, <span class=
```
mcdurdin, line 22:
```suggestion
<p>After compiling the keyboard successfully, we can start testing it. Choose <span class="guimenu">Debug</span>, <span class=
```

developer/15.0/guides/develop/tutorial/step-7.php
```diff
@@ -20,7 +20,7 @@
...
-"guimenuitem">Start Test</span> or press <kbd>F5</kbd> to begin testing. The Keyboard test window will appear.</p>
+"guimenuitem">Start Test</span> or press <kbd>F5</kbd> to begin testing. The Keyboard test window will appear.</p>
```
mcdurdin, line 23: "guimenuitem">Start Debugging</span> or press <kbd>F5</kbd> to begin testing. The Keyboard debug window will appear.</p>

mcdurdin (approved):

PR #624, opened 2022-10-27
docs: add reference for BCP 47

mcdurdin:
Fixes #571.

Also updates three files to markdown, but leaves remaining three as PHP for now -- they only had minor link changes.

PR #626, opened 2022-10-27
fix: lowercase link

darcywong00:
Follows #625

Lower-casing the metadata links from developer/15.0/reference also needs to be applied to developer/16.0/reference


mcdurdin (approved):

mcdurdin:
Fixed in #624 (b59eb77f6c13c82d7cb813b135c969e3aa899bc9)

PR #627, opened 2022-10-27
docs: kb0109 - blank keyboard on ios 16

sgschantz:
See issue keymanapp/keyman#7391 
This issue was first reported with FirstVoices, and Allow Full Access was enabled for FV with PR keymanapp/keyman#7459 to address it.


mcdurdin (approved):
LGTM. I fixed the issue links in the PR description :grin:

darcywong00:
This will need to 🍒 pick to master.  I presume you started on staging so you could see the page render locally.

mcdurdin:
> I presume you started on staging so you could see the page render locally.

Yes, this is a bit of a current gap if using Docker for local test at present. Once we use Docker/.htaccess throughout, this problem will disappear.

PR #646, opened 2022-11-08
docs(developer): update keyman.initialized docs

### LINKED ISSUE #644: docs: `keyman.initialized` flag can have value `2`
https://help.keyman.com/developer/engine/web/15.0/reference/core/initialized notes only values 0 and 1. However, 1 means init has been called but deferred init is not yet complete -- many things are not yet safe to call at this point. Value 2 means init is fully complete.

I'm not sure if we want to fixup the documentation or the property?

Also, it is still useful to be able to check the value of initialized from external scripts, so the text "Please note that this is included for legacy reasons from Keyman 1.0. Rather than blocking on this flag, it is safe to call keyman.init() instead to ensure initialization, so long as the page has already loaded." could be worded better to suggest using `keyman.init().then()` as a more appropriate alternative.

We should probably also make note of the fact that `keyman.init(null)` is automatically called on page load, so page authors should _not_ rely on this flag to know whether or not the initialization has completed with their preferred options...

mcdurdin:
Fixes #644


darcywong00 (approved):
lgtm

PR #675, opened 2022-12-09
Remove dev requirement for Composer

### LINKED ISSUE #674: bug: Update Docker build to eliminate dev local requirement of PHP 7.4 and Composer
Since the Docker image should already have PHP and Composer, we should be able to eliminate the
`./build.sh configure` step and not require the dev environment to have PHP 7.4 and Composer locally installed

@mcdurdin said he had to manually move the `/vendor` folder and edit his Dockerfile to copy Composer dependencies as
```
COPY --chown=www-data --from=builder /var/www/html/vendor /var/www/html/vendor
```

darcywong00:
fixes #674 

Since the Dockerfile has Composer, this separates the steps so the `composer install` step is handled in the Docker image (split into composer.dockerfile)

* There's also a minor refactor of the Windows path slashes into `_get_pwd()`
* Update readme.md with the IIS notes from https://github.com/keymanapp/keyman.com/issues/337#issuecomment-1336339387

TODO:
- [x] Update ci.yml to handle this separation


srl295 (commented):
looking good w some comments

PR #676, opened 2022-12-13
K8s convert pod to deployment

tim-eves:
We want this so a k8s cluster can automatically migrate a workload to another node, perform rolling upgrades and scale instances up or down. A bare pod is not really the right way to deploy outside prototyping.
Also
1. Reduce the number images we depend on for sidecar/init containers to just alpine, for a smaller foot print and smaller supply chain attack surface.
2. Move specification of which files in the repo to include/exclude in the document root out of the init container shell script and into a ConfigMap entry for easier maintenance and clarity.


darcywong00 (approved):
lgtm

PR #681, opened 2023-01-11
Remove dev requirement for Composer [Take 2]

### LINKED ISSUE #674: bug: Update Docker build to eliminate dev local requirement of PHP 7.4 and Composer

darcywong00:
Replaces #675

This is a second attempt to fix #674 and have the Docker file handle `composer install` so the dev environment doesn't need to have Composer installed.
Because of the way Docker handles filesystems, I had to move `build.sh configure` to link the `vendor/` folder after the container is running.

Previous attempts to copy `vendor/` to `/var/www/html/` don't work because the when the container starts, it replaces the folder with what's on the host (which doesn't have a `vendor/` folder)

Updates inspired from 
* [medium.com post](https://medium.com/@lukaspereyra8/docker-compose-php-composer-the-missing-vendors-folder-issue-66faa5475c59)
* [web-languageforge](https://github.com/sillsdev/web-languageforge/blob/ba4bf1ae838e48b4f2c2d7fa9861a7727eb04362/docker/app/Dockerfile#L38-L47)

## Steps for Developer
```bash
# Build the docker image and start the container
./build.sh build start 

# Link the vendor/ dependencies
./build.sh configure

# site now accessible at http://localhost:8055
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
         fail-fast: true
 
     #
-    # Install website PHP dependencies (copies vendor/ to the proper folder)
+    # Create link to vendor/ dependencies
     #
     - name:  Install dependencies
       shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,23 +22,22 @@ jobs:
       env:
         fail-fast: true
 
+    - name: Build the Docker image
+      shell: bash
+      # --mount option requires BuildKit
+      run: |
+        ./build.sh build start
+      env:
+        fail-fast: true
+
     #
-    # Install website PHP dependencies
+    # Install website PHP dependencies (copies vendor/ to the proper folder)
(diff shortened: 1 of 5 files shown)


darcywong00:
I note when switching branches, if the link to vendor is removed, the Dev will need to re-run
`./build.sh configure`

darcywong00:
The link-checker is also failing on the same ofl.txt pages when we've been merging master into stable each sprint.

srl295 (approved):

srl295:
@darcywong00  LGTM!

PR #685, opened 2023-01-20
fix: Move link of vendor/ from build.sh configure to start 🦭

darcywong00:
Follow-on to #681 
@mcdurdin pointed out `./build.sh configure` should occur before `./build.sh start`.

So this moves the link step from `configure` to the end of `start`.
This way Docker maps the local folder first, and creating the link to vendor/ stays.

Will propagate this change to other websites and update keymanapp/keyman#8034

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,22 +22,13 @@ jobs:
       env:
         fail-fast: true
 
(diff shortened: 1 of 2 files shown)


mcdurdin (approved):
LGTM

PR #688, opened 2023-01-25
Convert custom init git fetch script to use git-sync instead

tim-eves:
This is both shorter and less of a potential error magnet than a custom bash script.

### DIFF
--- a/resources/kubectl/help-kubectl.yaml
+++ b/resources/kubectl/help-kubectl.yaml
@@ -42,6 +42,8 @@ spec:
         volumeMounts:
         - name: help-site-app
           mountPath: /var/www/html
+          subPath: html
+          readOnly: true
         readinessProbe:
           httpGet:
             path: /robots.txt
@@ -55,6 +57,8 @@ spec:
       - name: api
         image: alpine
         imagePullPolicy: IfNotPresent
+        securityContext:
+          runAsGroup: 33
         env:
         - name: DEPLOY_KEY
           valueFrom:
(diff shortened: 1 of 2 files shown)


resources/kubectl/help-kubectl.yaml
```diff
@@ -85,42 +89,32 @@ spec:
...
-          echo Configure sparse-checkout
-          echo "$GIT_SPARSE_FILES" | git sparse-checkout set --no-cone --stdin
-          git checkout
-          echo Link vendored PHP
-          ln -sf /var/www/vendor vendor
+            configMapKeyRef: {name: help-keyman-com, key: site-branch}
+        args: [
+          --repo=https://github.com/keymanapp/help.keyman.com.git,
+          --sparse-checkout-file=/tmp/git/sparse-checkout.config,
+          --one-time, --depth=1, --root=/mnt, --dest=html]
```
darcywong00, line 105: So is `--branch=${SITE_GIT_BRANCH} \` no longer needed?

resources/kubectl/help-kubectl.yaml
```diff
@@ -85,42 +89,32 @@ spec:
...
+          mountPath: /mnt
+        - name: git-config
+          mountPath: /tmp/git
+
+      - name: init-site-vendoring
+        image: busybox
+        securityContext:
+          runAsUser: 0
+          runAsGroup: 33
+        command: ['sh', '-c', 'ln -sf /var/www/vendor /mnt/html/vendor']
```
darcywong00, line 117: l 44. has `mountPath: /var/www/html`. Does this link become
`ln -sf /var/www/vendor /mnt/var/html/vendor`?

resources/kubectl/help-kubectl.yaml
tim-eves: That option is for the `git` command the same info is passed via the `GIT_SYNC_BRANCH` env var in the git-sync based container, on line 99.

resources/kubectl/help-kubectl.yaml
tim-eves: The `mountPath` you mention is for the `app-php` contianer where as the `init-site-vendoring` container, where `ln` command is being run, is governed by the `mountPath: /mnt` on line 120.

darcywong00 (approved):
lgtm

PR #690, opened 2023-02-06
chore: Update current-version to 16.0

darcywong00:
@jahorton noticed https://help.keyman.com/DEVELOPER/ENGINE/WEB/current-version/ was still going to 15.0 content.

Now that 16.0 is released, this updates `current-version` from 15.0 to 16.0.

Note: staging branch will need the corresponding change in `.htaccess`

### DIFF
--- a/web.config
+++ b/web.config
@@ -45,7 +45,15 @@
                     <action type="Rewrite" url="{R:1}/XX.0{R:2}" />
                 </rule>-->
 
-                <rule name="current-version:15.0">
+                <rule name="current-version:16.0">
+                    <match url="^(.+)/current-version(/.*|$)" />
+                    <conditions>
+                        <add matchType="IsDirectory" input="{DOCUMENT_ROOT}/{R:1}/16.0" />
+                    </conditions>
+                    <action type="Rewrite" url="{R:1}/16.0{R:2}" />
+                </rule>
+
+                <rule name="version:15.0">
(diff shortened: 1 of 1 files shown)


mcdurdin:
@darcywong00 note the joyful capitalization in those URLs caused by a long-standing IIS bug. We'll need to handle that in our Apache-based version of the sites :-(

darcywong00:
> @darcywong00 note the joyful capitalization in those URLs caused by a long-standing IIS bug. We'll need to handle that in our Apache-based version of the sites :-(

ah, I was already handling /DEVELOPER in .htaccess (staging), but now I see I'll have to tweak for /DEVELOPER/ENGINE/WEB

https://github.com/keymanapp/help.keyman.com/blob/4909d73256ae9fc5a21e2332fc21e87c1757c712/.htaccess#L20-L21

mcdurdin:
> ah, I was already handling /DEVELOPER in .htaccess (staging), but now I see I'll have to tweak for /DEVELOPER/ENGINE/WEB

It's worse than that -- any path could have been full-capsed. However, I am wondering if it may be better to deliberately break those links, so that they'll disappear over time. Google's index will fix itself in a few days, and there will not be many other permalinks.

We could tweak the 404 error page to suggest the correct link in these cases?

darcywong00:
Capitalization issue will be addressed separately in #691

rc-swag (approved):
LGTM

PR #694, opened 2023-03-06
cask is now option, not command

Narvey:
When I tried to use cask as a command like this used to direct us to, I got the message:
    Error: `brew cask` is no longer a `brew` command. Use `brew <command> --cask` instead.


darcywong00 (approved):
lgtm (looks good to me).

I checked the Keyman for macOS install help, and it's also using the --cask option

https://github.com/keymanapp/keyman/blob/19b21e3bf6e5d2a589af06730fdbe1a62a550c6f/mac/help/start/install-keyman.md?plain=1#L37-L42

PR #705, opened 2023-04-24
chore: Fix typo in kmw.controlfocused.php

mcdurdin:


developer/engine/web/16.0/reference/events/kmw.controlfocused.php
```diff
@@ -24,7 +24,7 @@
 
 params("eventProperties", "object", "An object with the following properties:  "
   . "<ul><li><code>target</code> - the element gaining focus</li>"
-  . "<li><code>event</code> - the currently active control.</li>");
+  . "<li><code>activeControl</code> - the currently active control.</li>");
```
jahorton, line 27:
```suggestion
  . "<li><code>activeControl</code> - the previously-active control.</li>");
```

Since that's what we landed on for the intention.  I feel that to be a little clearer, since at the time the event is raised, `target` has gained focus and `activeControl` has lost it.

Unless I need to fix that too, anyway.

jahorton (approved):

PR #717, opened 2023-05-16
Fix various issues related to the api/deploy webhook

tim-eves:
Remove webhook match on non-existant keys in payload object. Fix issue with deployer script getting stuck, due to missing `--depth` flag, incorrect `git reset` invocation.
Add TLS object to ingress for dpeloyment manifest.


mcdurdin (approved):
LGTM

PR #731, opened 2023-06-18
feat(developer): add `&displayMap` reference 🗜

mcdurdin:
Relates to https://github.com/keymanapp/keyman/issues/9031


jahorton (approved):

DavidLRowe:
It seems that displaymap.md has "...the PUA range used should start at U+9100." but other text indicates that should be "...U+F100."

PR #737, opened 2023-06-30
chore: converts a second page with automation

jahorton:


developer/language/guide/variable-stores.md
```diff
@@ -0,0 +1,221 @@
...
+package allows keyboard options to be set during Keyboard Configuration,
+by selecting the Keyboard Options entry from the menu displayed by the
+Options button for the keyboard.
+
+Typically, the Options Configuration file will contain a heading, a form
+with radio buttons, or some other way of entering an option value, and
+the javascript needed to load the settings. A sample Option
+Configuration file is shown here:
+
+``` markup
```
jahorton, line 130: I do find it interesting that pandoc converted to this, rather than 

```suggestion
```html
```

... but it works all the same.

developer/language/guide/variable-stores.md
```diff
@@ -0,0 +1,221 @@
...
+            else
+              elem[j].value = val;
+        }
+      }
+    })();
+  </script>
+
+</body>
+</html>
+~~~
```
darcywong00, line 199: Should the code block end with? (ignore the > for formatting)
```suggestion
>```
```

jahorton: ... I must have gotten the regex wrong; thought both would have been found and replaced.  😦

developer/language/guide/variable-stores.md
```diff
@@ -0,0 +1,233 @@
...
+---
+
+Keyman 8 introduced support for dynamic variable stores, which can be
+utilized in many different ways. Some examples of applications of the
+use of variable stores are:
+
+-   Selection of alternate keyboard mappings, such as entering vowel
+    modifiers before or after the vowel characters.
+-   User control of whether to use <span class="key">CTRL</span> or
+    <span class="key">ALT</span> as a modifier key.
```
mcdurdin, line 12:
```suggestion
-   User control of whether to use <kbd>CTRL</kbd> or
    <kbd>ALT</kbd> as a modifier key.
```

This is probably outside scope but we'd be better using `<kbd>` now that it is available.

developer/language/guide/variable-stores.md
```diff
@@ -0,0 +1,233 @@
...
+somewhere in the keyboard source.
+
+## if() statement
+
+The [`if()` statement](../reference/if) allows any rule to be applied
+conditionally, according to whether or not the specified condition is
+true when the rule is to be used. The content of the store being used as
+a variable store can be compared with either a fixed string, or the
+content of another store. Note that since the
+[`match`](../reference/match) and [nomatch](../reference/nomatch) rules
```
mcdurdin, line 42:
```suggestion
[`match`](../reference/match) and [`nomatch`](../reference/nomatch) rules
```

developer/language/guide/variable-stores.md
```diff
@@ -0,0 +1,233 @@
...
+refer to all matched rules in a group, the `if()` statement cannot be
+used in a `match` or `nomatch` rule.
+
+The `if()` statement may be used in many different ways, such as:
+
+-   to provide alternate output for particular keystrokes:
+
+    ```
```
mcdurdin, line 50:
```suggestion
    ```keyman
```

developer/language/guide/variable-stores.md
```diff
@@ -0,0 +1,233 @@
...
+    store(opt1) '0'
+    if(opt1='1') + '$' > '£'
+    ```
+
+-   to conditionally branch to different rule groups, for example to
+    process pending context in different ways when the
+    <span class="key">spacebar</span> is pressed:
+
+    ```
```
mcdurdin, line 59:
```suggestion
    ```keyman
```

developer/language/guide/variable-stores.md
```diff
@@ -0,0 +1,233 @@
...
+    store(opt2) '0'
+    if(opt2 ='1') + " " > use(ruleGroup1)
+    if(opt2!='1') + " " > use(ruleGroup2)
+    ```
+
+-   to set or change option values depending on the value of an option,
+    or toggle between option settings:
+
+    ```
```
mcdurdin, line 68:
```suggestion
    ```keyman
```

developer/language/guide/variable-stores.md
```diff
@@ -0,0 +1,233 @@
...
+result of an entered key sequence. The `set` statement does not change
+the store permanently, and if Keyman is restarted, or Keyman
+Configuration modified, then the default value will be restored.
+
+In Keyman Desktop, changing a store with the `set()` statement while
+using one application will change it for all applications if the "Select
+keyboard layout for all applications" Keyman Configuration option is
+set.
+
+```
```
mcdurdin, line 93:
```suggestion
```keyman
```

developer/language/guide/variable-stores.md
```diff
@@ -0,0 +1,233 @@
...
+## save() statement
+
+The [`save()` statement](../reference/save) saves a current variable
+store value in system settings, to become the default value for future
+use. The `save()` statement will often be used following a `set()`
+statement to make a change permanent, as in the following example, which
+will, in effect, permanently change the content of store opt1 from `'0'`
+to `'1'`.
+
+```
```
mcdurdin, line 108:
```suggestion
```keyman
```

developer/language/guide/variable-stores.md
```diff
@@ -0,0 +1,233 @@
...
+
+## reset() statement
+
+The [`reset()` statement](../reference/reset) is used to restore a store
+to the current default value, either the value in the system settings
+from a previously matched rule using a `save()` statement, or else the
+store content as initialized in the keyboard source file.
+
+```
```
mcdurdin, line 120:
```suggestion
```keyman
```

developer/language/guide/variable-stores.md
```diff
@@ -0,0 +1,233 @@
...
+:   Tests a store or system state for a value
+
+[`reset()` statement](../reference/reset "reset() statement")
+:   Resets a store to its initial value
+
+[`save()` statement](../reference/save "save() statement")
+:   Persists a store value for next edit session
+
+[`set()` statement](../reference/set "set() statement")
+:   Sets a store or system state to a specific value
```
mcdurdin, line 233:
```suggestion
[Using stores](stores)
[`if()` statement](../reference/if)
:   Tests a store or system state for a value

[`reset()` statement](../reference/reset)
:   Resets a store to its initial value

[`save()` statement](../reference/save)
:   Persists a store value for next edit session

[`set()` statement](../reference/set)
:   Sets a store or system state to a specific value
```

mcdurdin (approved):
LGTM, a number of nits. I do think we need to consider if we can address these nits during the process, so that we don't end up with more maintenance in the future.

developer/language/guide/variable-stores.md
jahorton: Good news:  I've figured out how to handle this one automatically via Pandoc AST-parse filtering.

developer/language/guide/variable-stores.md
jahorton: I don't believe this is necessary.  If memory serves, because the page is underneath `/developer`, our syntax-highlighting engine is set with `keyman` as the default for the page's code blocks.

The converted page itself, as viewed within Docker, shows proper `keyman` highlighting without `keyman` here.

jahorton: It would be nice to find a way to configure a value here and set it automatically... but I don't see an easy way to pass a command-line argument to the filter setup that would allow us to insert it via automation.

developer/language/guide/variable-stores.md
jahorton: Seems like `pandoc` may struggle a bit on links that include code-styled text.

developer/language/guide/variable-stores.md
jahorton: Figured up a regex-matching approach for this.  It assumes that `<key>` elements should always be simple and not have nested elements.

PR #744, opened 2023-07-06
06-July-23 Converted stores.php to stores.md

Meng-Heng:
I deleted a language-keyman class in md since it was interfering with the whole page, will it be a problem in the future?

PR #747, opened 2023-07-07
chore: converts two more pages (no manual edits)

jahorton:


darcywong00:
See https://github.com/keymanapp/help.keyman.com/pull/744#issuecomment-1625454644 about CI link checker fails.

As the php files get converted to markdown, we should check to see if existing links to the php files also need to be updated.

jahorton:
> See [#744 (comment)](https://github.com/keymanapp/help.keyman.com/pull/744#issuecomment-1625454644) about CI link checker fails.
> 
> As the php files get converted to markdown, we should check to see if existing links to the php files also need to be updated.

That makes sense.  If only the link-checker actually told us what files were linking to them.

In the meantime, I just figured out a little more automation:  when converting a page, the automation tool will now strip file-extensions from internal links.  (`http://` and `https://` in a link will disable the 'stripping' behavior, while relative - and thus, internal - links will be sanitized.)

jahorton:
Also, the latest commit _is_ manual, but that's to patch up the busted links that crept in from recent PRs.

mcdurdin (approved):
LGTM

PR #750, opened 2023-07-14
Model to md

soysreyvy:


developer/lexical-models/submission/index.php
jahorton, line 1: This file should be deleted, now that we're replacing it.

darcywong00 (changes requested):
With the pending changes from #766, we'll want to remove `.php` from the 5 links

jahorton:
> With the pending changes from #766, we'll want to remove `.php` from the 5 links

That's the point of #766.  Rather than try to get the interns to update 20-odd PRs to resolve conflicts caused by merging it... I'll be merging these and then merge the result into #766; it's way easier to resolve things in one place, and the expected conflicts are all deleted files, which are notably easy to resolve in this case.

jahorton (approved):

PR #751, opened 2023-07-14
my commit

Meng-Heng:

PR #755, opened 2023-07-14
mengheng second commit of the day

Meng-Heng:


jahorton (commented):
The link-checker has noted that `visualkeyboard.php` tries to explicitly link to the PHP-based page `layoutfile.php`, which is converted by this PR.

I don't _think_ that `visualkeyboard.php` has been converted yet, so it'd probably be best to simply fix up its link to `layoutfile.php` (removing the `.php`) for now so that we don't introduce broken links.

----

Correction:  #751 has converted the offending page, though it does have some broken link checks of its own to resolve.  I'll look at this a bit more first and figure out how best to proceed.

jahorton (approved):

PR #756, opened 2023-07-14
mengheng third commit of the day

Meng-Heng:


developer/language/reference/nomatch.md
```diff
@@ -0,0 +1,62 @@
...
+nomatch > ...
+```
+
+## Description
+
+`nomatch` is similar to the [`match` rule](match), but instead of the
+rule being fired when a rule was matched, it will be used when no rule
+is matched in the group. A `nomatch` rule can include [`use()`
+statements](use), [`return` statements](return.php), [`beep`
+statements](beep.php) and characters.
```
jahorton, line 22:
```suggestion
statements](use), [`return` statements](return), [`beep`
statements](beep) and characters.
```

developer/language/reference/notany.md
```diff
@@ -0,0 +1,60 @@
...
+Support for `notany()` together with [`context()`](context) added in
+KeymanWeb 14.0.
+
+Introduced in version 7.0.
+
+## See also
+
+-   [Using stores](../guide/stores)
+-   [any() statement](./notany)
+-   [store() statement](./store)
```
jahorton, line 60:
```suggestion
-   [`any()` statement](./any)
-   [`store()` statement](./store)
```

To be fair, the `notany` part wasn't correct in the original version either... but we should fix it now that it's been noticed.

Pandoc didn't preserve code formatting on these links either.  That's sadly due to a tooling limitation.

jahorton (changes requested):

jahorton (approved):

PR #766, opened 2023-07-28
chore: removes .php file extensions from links of "current" pages

jahorton:
While the tool we're using for automatic page conversion from our PHP / HTML to markdown can remove link extensions from the files it converts, it doesn't handle inbound links from non-converted pages.  So, in the interest of avoiding a continuing string of even more broken links, this PR removes them, in-place, for the most relevant sections of the site.

What I did:

VS Code search (regex mode): `(href="[^"]+).php"` -> `$1"`

And then I filtered out the changes to older sections of the site (versions 15.0 and before).


ermshiperete (approved):
lgtm

jahorton:
Turns out I missed another fun case:

`(href=(["'])[^"']+).php(#[^"']+)["']` => `$1$3$2` - when a link has an included target-anchor name.

PR #770, opened 2023-07-28
Kl seven

kimleangkheng:


.gitattributes
jahorton, line 1: This one's going to be a bit trickier to fix; we don't want this file deleted - just left as it was.

jahorton:
Force-pushed so that the `.gitattributes` deletion doesn't show up in the commit history at all, since it was only within this branch.

jahorton (approved):

developer/16.0/context/debug.md
```diff
@@ -0,0 +1,175 @@
...
+Use Start Log/Stop Log to record the input and output. You can then use
+Run Test to run the test again, or go the Options menu to clear the log,
+or save or load a test, or use the batch mode to run several tests in a
+row.
+
+If the output produced while running a test is different to that stored
+when recording it, Keyman will halt the test on the line where the
+failure occurred, and activate Single Step mode.
+
+### <span id="id559616"></span>Status bar
```
jahorton, line 142: Just noticed this when looking back at the PR's changes - we have some empty `<span>` elements that came out of the PHP -> Markdown conversion.

PR #783, opened 2023-07-31
chore: remove shared lexical-model guide includes

jahorton:
Due to the Markdown conversion process, the content originally included from these files is now directly in each page variant that had been including it.  Markdown... doesn't really have "includes", so the change is kind of necessary.

This is being done based on #766 but is separate, as they're extra changes past what that PR is meant to address.

Draft until #766's link-checker pass comes up completely green.


darcywong00 (approved):
lgtm

PR #787, opened 2023-08-04
chore: removes old common _shared folder for developer/keyboards section

jahorton:
Both sections that previously source-included these pages have now been moved over to Markdown, so there's no longer anything referencing these files.


darcywong00 (approved):
lgtm

PR #793, opened 2023-08-04
Kl page

kimleangkheng:


developer/language/reference/ethnologuecode.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Description
+
+The optional `&EthnologueCode` store specifies the language the keyboard
+is made for by use of an ISO 639-3 three-letter language code. This
+information is intended for use by third-party applications, and is also
+used by the compiler to generate metadata for KeymanWeb.
+
+For a list of codes, or more information about Ethnologue, see
+[www.ethnologue.com](http://www.ethnologue.com/){.ulink
```
Nnyny, line 30:
```suggestion
[www.ethnologue.com](http://www.ethnologue.com/)
```

developer/language/reference/ethnologuecode.md
```diff
@@ -0,0 +1,67 @@
...
+target="_blank"}.
```
Nnyny, line 31:
```suggestion
.
```

developer/language/reference/includecodes.md
```diff
@@ -0,0 +1,82 @@
...
+
+### Parameters
+
+`filename`
+:   The file to import, relative to keyboard source path.
+
+## Description
+
+The `&includecodes` store allows you to reference an external file for
+[named character constants](../guide/constants){.link}. Keyman Developer
```
Nnyny, line 24:
```suggestion
[named character constants](../guide/constants). Keyman Developer
```

Nnyny (changes requested):

developer/language/reference/ethnologuecode.md
jahorton: Good catch!  The `target="_blank"` would admittedly be nice to keep... if it worked.  

I just tested adding that in to a different link on my machine... alas, I don't think we can use that feature.  So, good catch; we do need to remove both parts within the `{` and `}`.

jahorton (approved):

PR #801, opened 2023-08-22
fix: Use .localhost instead of .local 🦭

darcywong00:
This follows the sites-wide direction to use .localhost instead of .local for local Dev-hosted sites.
Mirrors keymanapp/keyman.com@3c9abe14fbb49d19bd3082081cfaf3ecf00cd82b

### DIFF
--- a/_common/KeymanHosts.php
+++ b/_common/KeymanHosts.php
@@ -119,7 +119,7 @@
         $this->r_keymanweb_com = "https://r.keymanweb.com";
       } else if($this->tier == KeymanHosts::TIER_DEVELOPMENT) {
         // Locally running sites via Docker need to access "host.docker.internal:[port]"
-        $this->s_keyman_com = "{$site_protocol}s.keyman.com{$site_suffix}";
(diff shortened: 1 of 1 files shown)


_common/KeymanHosts.php
```diff
@@ -109,12 +109,29 @@ function __construct() {
         $this->keyman_com = "https://keyman-staging.com";
         $this->keymanweb_com = "https://keymanweb.com";
         $this->r_keymanweb_com = "https://r.keymanweb.com";
+      } else if($this->tier == KeymanHosts::TIER_TEST) {
```
ermshiperete, line 112: It's unclear to me why we need this check and can't use the `else` branch below which seems to set the same values.

ermshiperete (approved):

_common/KeymanHosts.php
darcywong00: afaict, we'd use TIER_TEST on the CI GitHub actions, TIER_DEVELOPMENT for local dev environment, `else` for production. And each branch could diverge.

PR #816, opened 2023-09-26
refactor: split common parts out of build.sh

mcdurdin:


_common/docker.inc.sh
```diff
@@ -0,0 +1,107 @@
...
+  # Skip if link already exists
+  if [ ! -L vendor ]; then
+    # Create link to vendor/ folder
+    CONTAINER_ID=$(get_docker_container_id $CONTAINER_NAME)
+    if [ -z "$CONTAINER_ID" ]; then
+      builder_die "Docker container appears to have failed to start in order to create link to vendor/"
+    fi
+
+    docker exec -i $CONTAINER_ID sh -c "ln -s /var/www/vendor vendor && chown -R www-data:www-data vendor"
+  fi
```
darcywong00, line 104: Will docker.inc.sh eventually get refactored to a separate/shared repo? 
We would skip this for sites that don't use composer, right?

mcdurdin: > Will docker.inc.sh eventually get refactored to a separate/shared repo?

Yes, that's my thinking. We may have a code snippet to source it via curl, and then it will copy other shared elements in with `build.sh configure`

> We would skip this for sites that don't use composer, right?

Yes we may need to add a bit to skip the vendor copy for s.keyman.com etc.

darcywong00 (approved):
lgtm

PR #817, opened 2023-09-26
feat: move common files to keymanapp/shared-sites

mcdurdin:
This adds linkage to the new https://github.com/keymanapp/shared-sites repository, including downloading a bootstrap.inc.sh script from there to source all common files.

The first time build.sh is run, it detects that it needs to download and run the bootstrap. You'll see something like this in the log:

```
$ ./build.sh
[bootstrap] Bootstrap required
[bootstrap] Bootstrap starting
[bootstrap]   Downloading bootstrap.inc.sh
[bootstrap] Downloading _common files
[bootstrap]   Downloading _common/builder.inc.sh
[bootstrap]   Downloading _common/docker.inc.sh
[bootstrap]   Downloading _common/keyman-local-ports.inc.sh
[bootstrap]   Downloading _common/JsonApiFailure.php
[bootstrap]   Downloading _common/KeymanHosts.php
[bootstrap]   Downloading _common/KeymanSentry.php
[bootstrap]   Downloading _common/MarkdownHost.php
[bootstrap] All _common files downloaded
[bootstrap] Bootstrap complete
```

If you want to re-download, you can change the `BOOTSTRAP_VERSION` variable or call `./build.sh configure`. 

Caching note: raw.githubusercontent.com caches files on first retrieval. This means re-downloading files from github may get a stale version of a file if the `BOOTSTRAP_VERSION` variable is set to a branch or tag which has recently changed. Conversely, if it is set to a commit, then you are guaranteed to get the version in the commit, of course.

This goes together with https://github.com/keymanapp/shared-sites/pull/1. 

## TODO Before merge

Note that once https://github.com/keymanapp/shared-sites/pull/1 merges, we'll need to create a tag on shared-sites and refresh the version in this PR accordingly.

PR #820, opened 2023-09-29
fix: Add k8s step to pull shared-sites repo 🦭

darcywong00:
Follow-on to #817 which moved common website code to [[shared-sites](](https://github.com/keymanapp/shared-sites).
The k8s deploy step also needs to pull in this code (handled by `./build.sh configure`)


resources/kubectl/help-kubectl.yaml
```diff
@@ -109,6 +109,16 @@ spec:
...
 
+      - name: init-shared-sites-repo
+        image: k8s.gcr.io/git-sync/git-sync:v3.6.2
+        securityContext:
+          runAsUser: 0
+          runAsGroup: 33
+        command: ['sh', '-c', './build.sh configure']
+        volumeMounts:
+        - name: help-site-app
+          mountPath: /mnt
```
mcdurdin, line 120: - Place after init-site-vendoring
- Call it `configure-site`

mcdurdin (approved):

PR #825, opened 2023-10-15
chore: canonical keyboard version documentation

### LINKED ISSUE #19: [help.keyman.com] Hide deprecated keyboards from /keyboard index
From keymanapp/keyboards#252:

@LornaSIL writes:
> I would recommend that we remove deprecated keyboard help files from this page: http://https//help.keyman.com/keyboard/. The page is already a bit unwieldy and will get worse as we keep adding newer keyboards. The main (deprecated) keyboard page itself will still link to the deprecated help page.

@mcdurdin responds:
> To remove the deprecated items from the main index (or have them under a secondary index so that they can still be easily crawled if necessary), we'll need to do some work on help.keyman.com.

### LINKED ISSUE #824: chore: add rel=canonical and current-version links to keyboard documentation
Just as we have on the product documentation, we should have a similar pattern for keyboard documentation. This will help point search result at the latest version of documentation, addressing reports like: https://github.com/keymanapp/keyboards/pull/2395#issue-1937798912

mcdurdin:
* Fixes #824: adds a /keyboard/name/current-version canonical link
* Fixes #19: separates out deprecated keyboards and lists them later
* Also improves performance of keyboard index page, caching on first use.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -28,6 +28,13 @@ builder_describe \
 
 builder_parse "$@"
 
+function build_help_site() {
+  build_docker_container $HELP_IMAGE_NAME $HELP_CONTAINER_NAME
+  # if there is a cached keyboard index, remove it; it will be recreated on first use
+  # TODO: build this file during site build
+  rm -f keyboard/index.cache
+}
+
 function test_docker_container() {
   # TODO: lint tests
   composer check-docker-links
@@ -36,7 +43,7 @@ function test_docker_container() {
 builder_run_action configure  bootstrap_configure
 builder_run_action clean      clean_docker_container $HELP_IMAGE_NAME $HELP_CONTAINER_NAME
 builder_run_action stop       stop_docker_container  $HELP_IMAGE_NAME $HELP_CONTAINER_NAME
-builder_run_action build      build_docker_container $HELP_IMAGE_NAME $HELP_CONTAINER_NAME
+builder_run_action build      build_help_site
 builder_run_action start      start_docker_container $HELP_IMAGE_NAME $HELP_CONTAINER_NAME $HELP_CONTAINER_DESC $HOST_HELP_KEYMAN_COM $PORT_HELP_KEYMAN_COM
 
 builder_run_action test       test_docker_container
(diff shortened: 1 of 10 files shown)


build.sh
```diff
@@ -28,6 +28,13 @@ builder_describe \
 
 builder_parse "$@"
 
+function build_help_site() {
+  build_docker_container $HELP_IMAGE_NAME $HELP_CONTAINER_NAME
+  # if there is a cached keyboard index, remove it; it will be recreated on first use
+  # TODO: build this file during site build
```
darcywong00, line 34: We prob need @tim-eves to help with this TODO. I'm guessing in [help-kubectl.yaml](https://github.com/keymanapp/help.keyman.com/blob/staging/resources/kubectl/help-kubectl.yaml)?

keyboard/index.php
```diff
@@ -1,36 +1,49 @@
...
+        $items['deprecated'][$directory] = trim(substr($title, 0, $dep));
+      } else {
+        $items['active'][$directory] = $title;
+      }
     }
-    $items[$directory] = $title;
+
+    file_put_contents(INDEX_CACHE, json_encode($items));
+  } else {
+    $items = json_decode(file_get_contents(INDEX_CACHE), true);
```
darcywong00, line 44: How does `INDEX_CACHE` get update when new keyboards are deployed (w/o rebuilding the site)

build.sh
mcdurdin: No, it goes right here in build.sh. It's low priority because it's built on first use anyway.

keyboard/index.php
mcdurdin: The site gets rebuilt when it's redeployed, doesn't it?

darcywong00: ah right.

darcywong00 (approved):
lgtm

build.sh
mcdurdin: We'll move the cache build step to `configure` step, so it will be triggered by deploys in k8s

PR #827, opened 2023-10-16
chore: do build.sh configure in docker action

mcdurdin:
- Removes build.sh configure from k8s step
- Adds it to docker action instead
- Also fixes the image for k8s app-php


darcywong00 (approved):
lgtm

PR #830, opened 2023-10-16
chore: Set line endings for .sh files

darcywong00:
I was trying to run the help site on a Windows host machine and ran into
```bash
/usr/bin/env: 'bash\r': No such file or directory
```

I think this got it working for me.


mcdurdin:
I think this may break things in docker though?

darcywong00:
The docker image is based on ubuntu, so it's already using lf line endings, right?

mcdurdin (approved):

PR #849, opened 2023-10-26
chore: also merge if 'behind'

mcdurdin:
If we are ready to auto-merge, but the branch is behind master, we should still auto-merge if there are no conflicts


mcdurdin:
Not required -- I removed the branch protection rule requiring up-to-date because it is probably overkill on this repo.

PR #867, opened 2023-11-08
chore: Update "Key text" table for keymanweb-osk.ttf v4.0

darcywong00:
Documents the non-printing characters added in 17.0 to:
* Developer keymanapp/keyman#9846 
* KeymanWeb keymanapp/keyman#9547 (keymanweb-osk.ttf) 
 
with the [font-symchar v4.0](https://github.com/silnrsi/font-symchar/blob/main/documentation/encoding.md).

At some point, maybe we move this entire table out of the guide into a kb article? I find it very hard to locate the special symbols :confused:


mcdurdin (approved):
LGTM

> At some point, maybe we move this entire table out of the guide into a kb article? I find it very hard to locate the special symbols 😕

Agree that it's hard to find. The table is reference documentation; it doesn't belong in a KB article, which are more about issues and workarounds, but it should be in a reference document rather than a chatty tutorial topic with the title "nitty gritty". We don't currently have a good reference document for the touch layout data, so that'd be a starting point, probably under /developer/language/guide I think.

PR #869, opened 2023-11-08
chore: Merge master to staging for Sprint A17S25

darcywong00:
End of sprint task

Though I wonder if we disable the auto-merge workflow on staging?


mcdurdin (approved):
rslgtm

darcywong00:
> The workflow will essentially be ignored on staging because keyman-server is not creating PRs that base on staging. So I think it won't be an issue.

gotcha -

PR #877, opened 2023-11-15
chore: provide tier info in _control/info 🦭

mcdurdin:
Relates to keymanapp/keyman.com#403


darcywong00:
Would we want this in the shared-sites repo?

mcdurdin:
> Would we want this in the shared-sites repo?

I considered that but thought that we'd have different data in some sites -- it's so trivial that I am not too concerned.

e.g. I will move api.keyman.com/index.php to api.keyman.com/_control/info eventually.

darcywong00 (approved):
lgtm

PR #903, opened 2023-12-04

keyman-server:


mcdurdin:
Closing this, will be replaced when fix in keymanapp/keyboards#2483 lands.

PR #952, opened 2024-01-24
chore: 4 files markdown conversion 'developer/16.0/guides/distribute/[files]'

Meng-Heng:
The [packages](https://help.keyman.com/developer/16.0/guides/distribute/packages) page is missing a style for the **Tip** section after the conversion.
There is a [workaround ](https://github.com/orgs/community/discussions/16925)by using <kbd>></kbd> to indent the section. 

@mcdurdin, I only pushed 8 files for this request. Are the extra files from the bot or did I not pull correctly?


mcdurdin:
> @mcdurdin, I only pushed 8 files for this request. Are the extra files from the bot or did I not pull correctly?

It looks like you might have based your local branch on `master`, so you've pulled in changes on `master` that are not yet on `staging`. There are two ways to address this:

1. Rebase your local branch on `staging` (perhaps using an interactive rebase to pick only the commits you want), and then force-push to this PR branch, or
2. We update `staging` with changes from `master`, then update this PR to bring it up to date with `staging` -- this will cause the additional, unrelated commits to drop out of this PR branch.

Option 1 is faster because it doesn't take coordination with other team members, but option 2 is probably better!

mcdurdin:
You'll see I went with option 2. See #953. If you compare the commits listed you can see how they match up with your PR's commits (compare particularly the commit hashes on the right hand side):

image

So now, we'll get #953 approved and merged, and then click Update branch on this PR, and those commits should disappear from here!

mcdurdin:
> There is a [workaround ](https://github.com/orgs/community/discussions/16925)by using > to indent the section.

I've added an issue #954 to track this.

mcdurdin (approved):
LGTM

mcdurdin:
Note: we need to copy these changes to the 17.0 folder as well

PR #961, opened 2024-01-31
fix: replace old-style lang codes, matching actual use

jahorton:
Fixes a mismatch between the language codes actually used by an example and what the example _says_ the codes are.  Obviously, in favor of the former.

### DIFF
--- a/developer/engine/web/17.0/guide/get-started.php
+++ b/developer/engine/web/17.0/guide/get-started.php
@@ -49,8 +49,8 @@ echo codebox(<<<END
   <script>
     (function() {
       keyman.init({attachType:'auto'}).then(function() {
-        keyman.addKeyboards('@eng'); // Loads default English keyboard from Keyman Cloud (CDN)
-        keyman.addKeyboards('@tha'); // Loads default Thai keyboard from Keyman Cloud (CDN)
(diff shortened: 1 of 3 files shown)


developer/engine/web/17.0/guide/get-started.php
```diff
@@ -49,8 +49,8 @@
   <script>
     (function() {
       keyman.init({attachType:'auto'}).then(function() {
-        keyman.addKeyboards('@eng'); // Loads default English keyboard from Keyman Cloud (CDN)
-        keyman.addKeyboards('@tha'); // Loads default Thai keyboard from Keyman Cloud (CDN)
+        keyman.addKeyboards('@en'); // Loads default English keyboard from Keyman Cloud (CDN)
+        keyman.addKeyboards('@th'); // Loads default Thai keyboard from Keyman Cloud (CDN)
```
jahorton, line 53: Contrast:

https://github.com/keymanapp/help.keyman.com/blob/910c5eb0b53c848a088141ebb99645b531fc9487/developer/engine/web/17.0/guide/get-started.php#L35

https://github.com/keymanapp/help.keyman.com/blob/910c5eb0b53c848a088141ebb99645b531fc9487/developer/engine/web/17.0/guide/examples/__first-example.php#L15-L22

darcywong00 (approved):
lgtm

I think normally we just bother updating the latest stable -onward

PR #967, opened 2024-02-01
chore: convert 'developer/16.0/guides/lexical-models/advanced/...'

Meng-Heng:
Location of conversion - https://help.keyman.com/developer/16.0/guides/lexical-models/advanced

The order of the content in the left sidebar after conversion has been switched automatically.
image

### DIFF
--- a/developer/16.0/guides/lexical-models/advanced/word-breaker.md
+++ b/developer/16.0/guides/lexical-models/advanced/word-breaker.md
@@ -6,16 +6,14 @@ The `trie` family of lexical models needs to know what a word is in
 running text. In languages using the Latin script—like, English, French,
 and SENĆOŦEN—finding words is easy. Words are separated by spaces or
 punctuation. The actual rules for where to find words can get quite
-tricky to describe, but Keyman implements the [Unicode Standard Annex
-\#29 §4.1 Default Word Boundary
-Specification](https://unicode.org/reports/tr29/#Word_Boundaries) which
-works well for most languages.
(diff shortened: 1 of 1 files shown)


darcywong00:
Thanks for updating these. Since some .php files are converted to .md, there's a few more links to update:
/developer/16.0/guides/lexical-models/advanced/unicode-breaker-extension.md has a link
```
The word breaker function can be specified in the
[model definition file](./model-definition-file.php) as follows:
```
that needs to just be `./model-definition-file`

I wonder if the developer/17.0/ content also needs to be migrated too.

Meng-Heng:
Thank you, Darcy. 
I modified the unicode-breaker-extension.md

mcdurdin:
> I wonder if the developer/17.0/ content also needs to be migrated too.

Yes. At this point it's probably possible to copy the .md from 16.0 but careful diffing should be done first. I'm a little swamped to take this on!

developer/16.0/guides/lexical-models/advanced/model-definition-file.md
```diff
@@ -0,0 +1,107 @@
...
+};
+export default source;
+```
+
+Let's step through this file, line-by-line.
+
+On the first line, we're declaring the source code of a new lexical
+model.
+
+``` lang-typescript
```
darcywong00, line 45: I think the Markdown syntax highlighting uses this

```suggestion
```typescript
```

Applies to all the other typescript code blocks on the PR

Meng-Heng (commented):
remove lang-typescript and cleanup files

darcywong00 (approved):
lgtm

PR #971, opened 2024-02-02
chore: convert 'developer/16.0/guides/lexical-models/tutorial/...'

Meng-Heng:
Hello Darcy and Marc,

The figure tag is present in step-4. Is it okay to use that tag?

As for step-2, I made a bunch of changes to the figcaption and figure tags since it did not position them correctly when converted.

in step-3, some changes weren't possible:

- table center alignment with its text
- one of the img is way bigger and can't resize (it is possible to add width if this img is wrap in a figure tag like step-4)

Let me know if the files need any more changes.


developer/16.0/guides/lexical-models/tutorial/step-3.md
```diff
@@ -0,0 +1,152 @@
...
+the words in the language, with each word being separated by a line
+break.
+
+This is what my word list looks like in Google Sheets:
+
+<div markdown="1" style="text-align: center">
+
+![screenshot of the word list in Google
+Sheets](/cdn/dev/img/developer/120/lm/sencoten-sheets-full.png)
+
```
darcywong00, line 129: Hm, this image looks cropped (though the file is 1678 pixels wide)

Compare to the live site
https://help.keyman.com/developer/16.0/guides/lexical-models/tutorial/step-3#toc-editing-the-tsv-in-google-sheets
where the image is rescaled to 740 pixels wide.

Might need @mcdurdin suggestions how to style this. Or if we resize the `sencoten-sheets-full.png` file.

mcdurdin: I think this is a separate issue of styling images within markdown. I think we'll setup a max-width and scaling for that in CSS. Shouldn't block this change.

PR #972, opened 2024-02-02
chore: convert 'developer/16.0/guides/lexical-models/intro/index' & 'developer/16.0/guides/lexical-models/index'

Meng-Heng:
All the figure and figcaption tags are changed to img tags which then got converted to markdown image tags.


Meng-Heng (commented):
I made changes to span and kbd tags (/intro/index.md)

darcywong00 (approved):
lgtm

PR #985, opened 2024-02-06
chore: files cleanup in '16/reference/file-types'

Meng-Heng:
I tried rebasing before the conversion but the file from 'developer/16.0/reference/tools/index.php' (#981) reappears in every commit. I didn't see the file when using `git rebase -i staging` so I was unable to remove it.


darcywong00 (approved):
lgtm

PR #996, opened 2024-02-09
docs(linux): Add KB article troubleshooting typing problems on Linux

ermshiperete:


darcywong00:
Oh, can we add a link to this at the bottom of knowledge-base/index.php?

PR #1009, opened 2024-02-14
chore: Update ofl files to fix broken links

darcywong00:
Attempting to fix CI fails on keyboard deployment because of broken links in the legacy/ipauni11/OFL*.txt
See https://github.com/keymanapp/help.keyman.com/actions/runs/7891872735/job/21537118035?pr=1007

## Reference
https://openfontlicense.org/ofl-faq/

Not updating keyboard source since it's a deprecated keyboard


mcdurdin:
I think we replace this with #1011, as I don't believe links in .txt files are followed, and in any case external links are definitely not followed.

darcywong00:
Replaced with #1011

PR #1011, opened 2024-02-14
chore: no redirect for 404

mcdurdin:
The .php extension on the 404 ErrorDocument directive was causing Apache to do a redirect. Removing this means that the original url is maintained.

Also did a little bit of cleanup to the error page at the same time:
* removed (unhelpful) breadcrumbs
* gave a direct link to Keyman Community
* removed IIS-specific code

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -14,7 +14,7 @@ RewriteEngine on
 RewriteBase /
 
 # Custom error messages (need to pass the original request)
-ErrorDocument 404 /_includes/errors/404.php?uri=%{REQUEST_URI}
+ErrorDocument 404 /_includes/errors/404
 
 # Handle IIS bug which created links to /DEVELOPER instead of /developer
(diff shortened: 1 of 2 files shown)


darcywong00 (approved):
lgtm

darcywong00:
hmm, CI failed because aramaic/1.1/aramaic.php refers to aramaic.ico 
even though #625 changed the linke to Aramaic.ico (the actual filename)

mcdurdin:
> The current line has history from 2 years ago, before the change 😕

Yeah because the keyboard deployment keeps on overwriting the file. See https://github.com/keymanapp/keyboards/blob/master/legacy/a/aramaic/source/help/aramaic.php#L103

mcdurdin:
https://github.com/keymanapp/keyboards/pull/2555 should fix the broken ico link

PR #1086, opened 2024-03-05
docs(developer): tweak content around fonts

### LINKED ISSUE #895: docs(developer): how to add fonts to the keyboard package
There is no video or document on how to add a font to a keyboard (package). The resource we found when searching on this issue is on this help page: https://help.keyman.com/DEVELOPER/16.0/guides/distribute/tutorial/step-2.

Here is the section on Font, but there is no instructions to add it to the package. What is mentioned however is to create the files for the keyboard package. The thing is we do not create the font, and we only want to include it in there.

image

mcdurdin:
Fixes #895.

The documentation needs more comprehensive work; this is addressing a single issue around creating vs using fonts, with minor tweaks on the same page.

### DIFF
--- a/developer/17.0/guides/distribute/tutorial/step-2.php
+++ b/developer/17.0/guides/distribute/tutorial/step-2.php
@@ -8,7 +8,7 @@
 
 <h1>Step 2: Creating Additional Files for the Package</h1>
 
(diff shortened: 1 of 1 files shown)


SabineSIL (approved):
LGTM

PR #1134, opened 2024-03-21
chore: support `latest-version` url format

mcdurdin:
This aligns with the work on kmn.sh for error message redirects from kmc.

See also keymanapp/keyman#10888.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -53,83 +53,92 @@ RewriteRule "^keyboard/(?!redir.php)([^/?]+)/([^/?]+)(/)?$" "keyboard/redir.php?
 # RewriteCond "%{DOCUMENT_ROOT}/$1/XX.0" -d
 # RewriteRule "^(.+)/current-version(/.*)?$" "$1/XX.0$2"
 
-
 # Redirect unversioned paths to current-version/ path
 
 # /products first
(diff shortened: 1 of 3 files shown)


.htaccess
```diff
@@ -53,83 +53,92 @@ RewriteRule "^keyboard/(?!redir.php)([^/?]+)/([^/?]+)(/)?$" "keyboard/redir.php?
...
+
+# latest-version:18.0
+# alpha or beta version reference
+RewriteCond "%{DOCUMENT_ROOT}/$1/18.0" -d
+RewriteRule "^(.+)/latest-version(/.*)?$" "$1/18.0$2" [DPI]
+
+# latest-version:17.0
+# alpha or beta version reference
+RewriteCond "%{DOCUMENT_ROOT}/$1/17.0" -d
+RewriteRule "^(.+)/latest-version(/.*)?$" "$1/17.0$2" [DPI]
```
darcywong00, line 75: What's the distinction between latest-version and current-version? Is current-version the current stable version?

mcdurdin: yes, `latest-version` is the most recent alpha release (e.g. 18.0 at this time) that can be found, whereas `current-version` is always going to be latest stable

mcdurdin: probably should have gone with `alpha` and `stable` as tags, still could do that moving forward!

darcywong00 (approved):
lgtm

Issue #1089, opened 2024-03-05
docs(developer): New Project Parameters, Package Editor need doc updates for 17.0 fields


mcdurdin:
e.g. Description is new to 17.0, plus a number of others.

mcdurdin:
Blocked on #1107, #1110, and any follow-ups on those

PR #1147, opened 2024-03-25
chore: 17.0 context documentation updates

Meng-Heng:
Hello!
Some images are not in use for the files here but they are in #1148 PR. Please approve this PR before #1148.


darcywong00:
I wonder since some of the images are in a 170 folder, maybe can drop the "17.0" in the filename

For example:
170/ui/Debug_Menu17.0.png could be Debug_Menu.png

Meng-Heng:
It uses the 10.0 images when I use Debug_Menu and it does for other images as well. Should I instead use Debug_Menu_1 after the name?

darcywong00:
> Edit: Darcy, I got it working now.

Nice!  I think there's at least 5 more image filenames this applies to

Meng-Heng:
Thank you, Darcy. I got them all renamed now.

darcywong00 (approved):
lgtm

mcdurdin:
I think this fixes #1089, do you agree?

PR #1209, opened 2024-04-24
fix: Remove html entities in keyboard pagenames

darcywong00:
I noticed for keyboard pages that have special characters in pagename (e.g.)
* devu-mikes
* sil_nko
The special characters in `$pagename` are already use htmlentities.

The page titles end up with the literal string (e.g. `DevnagriUnicode Mike&apos;s Keyboard Help`
https://help.keyman.com/keyboard/devu-mikes/1.0a/devu-mikes

This fixes the page titles.


ermshiperete (approved):

mcdurdin:
I don't think this fix is appropriate. We have other pages which will break if we do this. The correct fix is to take the HTML entities out of the $pagetitle in the affected pages in keyboards.

darcywong00:
> I don't think this fix is appropriate. We have other pages which will break if we do this. The correct fix is to take the HTML entities out of the $pagetitle in the affected pages in keyboards.

Are you saying the keyboard pages would update pagetitle so
`DevnagriUnicode Mike&apos;s Keyboard Help` becomse
`DevnagriUnicode Mike\'s Keyboard Help`?

mcdurdin:
> Are you saying the keyboard pages would update pagetitle so
> `DevnagriUnicode Mike&apos;s Keyboard Help` becomse
> `DevnagriUnicode Mike\'s Keyboard Help`?

Yes

darcywong00:
Back to draft. I may fix up the older versions of the keyboard help. But will update the keyboards repo first.

mcdurdin (approved):

PR #1238, opened 2024-05-07
chore: Update kb0062.md

Nnyny:


knowledge-base/kb0062.md
```diff
@@ -1,7 +1,5 @@
 # HOWTO: Find fonts that work with your Keyman Desktop keyboard
```
darcywong00, line 1: I thought we were leaving some of these old KB articles as "archived documentation".

If we are updating them, maybe rename "Keyman Desktop" to "Keyman for Windows" on this and other articles as you update...

Nnyny: Noted, @darcywong00.

jahorton (approved):

PR #1265, opened 2024-05-16
fix: keyman.options has become keyman.config.options

mcdurdin:
`keyman.options` was not a published API so there was no guarantee of stability. It changed to `keyman.config.options` in 17.0. Note: this is not a published API either.

Fixes: keymanapp/keyman#11466
Fixes: HELP-KEYMAN-COM-BWG


jahorton:
This is not the right way to fix it:  https://help.keyman.com/developer/engine/web/17.0/reference/util/getOption is the stable API point.

PR #1271, opened 2024-05-17
fix: should use published API for KMW option access

jahorton:
Fixes https://github.com/keymanapp/keyman/issues/11466.
Reworks #1265.

Uses a pre-existing published API: https://help.keyman.com/developer/engine/web/17.0/reference/util/getOption


mcdurdin (approved):

mcdurdin:
hmm, seeing the other utility functions there, `util` feels like the wrong place for that function, but that's a separate discussion. :grin:

PR #1272, opened 2024-05-17
change: use published Keyman Engine API to fetch keyboards in keyboard help pages

jahorton:
While fixing https://github.com/keymanapp/keyman/issues/11467, I noticed that some of the font-styling for keyboards wasn't being applied.  It turns out that this was due to help.keyman.com bypassing keyboard stubs (Web's equivalent to kmp.json) entirely, thus breaking the link between keyboard and fonts.

In older versions of Keyman Engine for Web, this likely wasn't an issue due to how entangled OSK code was with the rest of the Keyman Engine implementation.  Part of preparing the OSK to (eventually) become a standalone module means it needs to rely on data that the OSK, by itself, has access to - and the design for that is reliant upon keyboard stub information.


mcdurdin:
> In older versions of Keyman Engine for Web, this likely wasn't an issue due to how entangled OSK code was with the rest of the Keyman Engine implementation. Part of preparing the OSK to (eventually) become a standalone module means it needs to rely on data that the OSK, by itself, has access to - and the design for that is reliant upon keyboard stub information.

Note that the original design for KeymanWeb supported direct reference to keyboards with script elements rather than stubs. So if this no longer works, we really need to clearly document this as a breaking change. It may impact other users.

mcdurdin:
> To the best of my knowledge, _pre-loading_ the keyboards was intended - but there was nigh-always a stub loader paired with it as well.

Not sure why you are arguing with the person who did the original design.

In the original design, the intent of the keyboard stub was to avoid the network cost of loading a keyboard that wouldn't be needed. The stub was not originally intended to have metadata that the keyboard itself didn't have, and so was optional. (Not saying this is good design, just saying what it was.) At the time of the original design that XMLHttpRequest/AJAX didn't exist. We can do better now of course.

cdn/dev/js/kbd-docs.js
```diff
@@ -74,8 +74,84 @@ function loaded(){
...
     var url = location.pathname.split('/');
-    var keyboardName = url[url.length-3], keyboardVersion = url[url.length-2];
-    var keyboardPath = keyman.util.getOption('keyboards') + keyboardName + '/' + keyboardVersion + '/' + keyboardName + '-' + keyboardVersion + '.js';
+    var keyboardName = url[url.length-3]; //, keyboardVersion = url[url.length-2];
+
+    // Ensure that the Web engine itself fetches the cloud metadata about the keyboard
+    // so that the OSK knows about the fonts.
+    keyman.addKeyboards(keyboardName).then(function () {
+      // Ensure that the keyboard has been fetched.
+      return keyman.setActiveKeyboard(keyboardName, '');
```
mcdurdin, line 83: We don't want to activate the keyboard -- it's only used for documentation.

cdn/dev/js/kbd-docs.js
```diff
@@ -95,6 +171,8 @@ function loaded(){
       return null;
     }
 
+    // Even with KMW loading the fonts for itself, we wish for them to be available on all page elements,
+    // not just in Web's OSK.  We _could_ use the retrieved stub to get that, but... "ain't broke, don't fix it."
```
mcdurdin, line 175:
```suggestion
    // Even with KMW loading the fonts for itself, we wish for them to be available on all page elements,
    // not just in Web's OSK. 
    // TODO: consider using the retrieved stub to get that in future
```

mcdurdin (changes requested):
I think we need to look at alternatives for the setActiveKeyboard call

cdn/dev/js/kbd-docs.js
jahorton: This is the public API mechanism Web uses to actually fetch the keyboard.

We could use the unofficial, internal `keyman.keyboardCache.fetchKeyboard()` to accomplish this instead, though note that it's KMW 17.0+ only.

jahorton: Finally, there is the possibility of just... not removing the direct-load of the JS file (as before) while still requesting the stub from the cloud (as added).  It'd take a bit of "`Promise` glue" to ensure the order-of-operations is carried out in proper order.

mcdurdin: OK, that's very side-effecty! We should have a public API for fetching the keyboard.

For now I am okay with using `fetchKeyboard` here so long as we publish an API before 18.0 stable.

Note that I've rolled the help site back to 16.0 temporarily (#1278), a patch we should remove when we merge this PR.

jahorton:
Double-checking... yep, we're in business.  The current stable-17.0 works properly with these changes.

image

cdn/dev/js/kbd-docs.js
jahorton: I've merged in `master` and have added a commit to undo the patch on this PR, if that's acceptable.

jahorton: > OK, that's very side-effecty! We should have a public API for fetching the keyboard.
> 
> For now I am okay with using `fetchKeyboard` here so long as we publish an API before 18.0 stable.

Documented as https://github.com/keymanapp/keyman/issues/12102.

jahorton:
With 16.0 in place...

image

Differences:
- the missing left-hand padding is the same for both.
- 16.0 lacks the keyboard name in spacebar text (as opposed to having it, but at too large a scale)
- 16.0 does not affect page text, whereas 17.0 does
- 16.0 cannot display the layout's hints, as those are 17.0 only.

mcdurdin:
> * 16.0 lacks the keyboard name in spacebar text (as opposed to having it, but at too large a scale)
> * 16.0 does not affect page text, whereas 17.0 does

These two, particularly the second one, seem like they are pretty important to fix before merge.

> * the missing left-hand padding is the same for both.

This could be resolved separately.

This seems to me to be three separate issues that need to be opened for KeymanWeb.

jahorton:
> > * 16.0 does not affect page text, whereas 17.0 does
> 
> These two, particularly the second one, seem like they are pretty important to fix before merge.

Digging into this point:  the `naijatype` keyboard has the following in its associated CSS:

```css
body {
    background: rgb(193, 194, 193);
    font-family: "Andika","DejaVu Sans","AndikaAfr","Andika Afr","Andika New Basic Compact";
}
```

https://github.com/keymanapp/keyboards/blob/03b932e9b9558b6885534d31934084d00e88a6ef/experimental/n/naijatype/source/naijatype.css#L6-L9

The "affecting page text" issue is a consequence of this - we wish to preserve OSK styling, and there is some actual OSK-styling content in that CSS file.  It's just that some of it is decidedly _not_ targeted for the OSK.

jahorton:
Things look fine for a number of other keyboards; I think the observed issues are due to interactions with that keyboard's custom CSS.

mcdurdin:
> Digging into this point: the `naijatype` keyboard has the following in its associated CSS:

I've opened keymanapp/keyboards#3039 to address this.

cdn/dev/js/kbd-docs.js
```diff
@@ -74,8 +74,87 @@ function loaded(){
...
+    var keyboardName = url[url.length-3]; //, keyboardVersion = url[url.length-2];
+
+    // Ensure that the Web engine itself fetches the cloud metadata about the keyboard
+    // so that the OSK knows about the fonts.
+    keyman.addKeyboards(keyboardName).then(function () {
+      // Ensure that the keyboard has been fetched.
+      //
+      // Note:  INTERNAL API - not documented.  This loads the keyboard (given
+      // the `addKeyboards` call), but does not _activate_ it.
+      return keyman.keyboardRequisitioner.cache.fetchKeyboard(keyboardName);
```
jahorton, line 86: Per [review request](https://github.com/keymanapp/help.keyman.com/pull/1272#discussion_r1606289845), `fetchKeyboard` has replaced previous use of `setActiveKeyboard`.

jahorton (commented):
What's the state of this PR now?  I believe I've addressed the main change to be requested, I've got a fix up for the spacebar-label overflow issue, and the other big item appears to be specific to a single keyboard.

Is there anything else I might have overlooked here?  Granted, we _may_ want to wait for the spacebar-label to land on stable.

mcdurdin (approved):
LGTM

Issue #1058, opened 2024-02-26
chore: include the version switcher in the 404 page


mcdurdin:
![image](https://github.com/keymanapp/help.keyman.com/assets/4498365/280c74bb-6c2f-4d2b-bce6-c7733f26a5c7)

It may be that the current-version is a release version which doesn't yet have the feature, but the upcoming release will have the feature. So we should (a) include the version switcher, and (b) offer to take the user to the pre-release content.

For example:

![image](https://github.com/keymanapp/help.keyman.com/assets/4498365/551e6b2d-1777-4666-891c-6e23274d740a)

mcdurdin:
> (b) offer to take the user to the pre-release content.
For example, if current-version is 16.0, and we navigate to /developer/current-version/reference/api/kmc-analyze, then we could look at /developer/latest-version/reference/api/kmc-analyze, see a file there (kmc-analyze.md) and add a link to the 404 page to take them there.

PR #1280, opened 2024-05-21
feat: include the version switcher in the errors page and show where the URL's subpath is located across the versions' folders

Meng-Heng:
Fixes: #1058

Case 1:
image

Case 2:
image

The content can only be found if the URL is using the structure: basePath + versionedPath + subPath. If found, the version switcher will show as the Version. If not, it will direct you to the home page. 

Please kindly review.

### DIFF
--- a/_includes/includes/page-version.php
+++ b/_includes/includes/page-version.php
@@ -60,22 +60,18 @@
       }
 
       // Build the menu
-      echo "<div id='version-selector'>Other versions";
-      echo "<div>";
+      $html = "<div id='version-selector'>Other versions";
+      $html .= "<div>";
+
+      $found = false;
 
       foreach($PageVersion->versions as $version) {
-        $found = true;
         $folder = "{$_SERVER['DOCUMENT_ROOT']}{$PageVersion->basePath}/{$version}/";
         if(!file_exists("{$folder}{$file}.php") && !file_exists("{$folder}/{$file}.md")) {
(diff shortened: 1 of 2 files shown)


_includes/includes/page-version.php
```diff
@@ -64,16 +64,17 @@ static function WriteSelector() {
...
 
       foreach($PageVersion->versions as $version) {
+        $found = true;
         $folder = "{$_SERVER['DOCUMENT_ROOT']}{$PageVersion->basePath}/{$version}/";
         if(!file_exists("{$folder}{$file}.php") && !file_exists("{$folder}/{$file}.md")) {
           // We don't have the same content in that version, so point to home page
           if($version == $PageVersion->currentVersion) {
             $text = "Version $version (home page, current version)";
-            $target = "{$PageVersion->basePath}/current-version/";
+            $target = "{$PageVersion->basePath}/current-version/{$PageVersion->subPath}";
```
mcdurdin:
```suggestion
            $target = "{$PageVersion->basePath}/current-version/";
```

I don't think this change should be here, either :grin:

mcdurdin (approved):
I've pushed a commit which changes page-version a bit, to only show versions which include content. You might like to have a look at the changes I made and see if they make sense to you. 😇

Meng-Heng:
It makes sense to me. 
There aren't any pages to point to home for each version, in a way it looks cleaner.
image
image

PR #1284, opened 2024-05-23
chore: update image of virtual key identifier 

Nnyny:


mcdurdin (approved):
LGTM; One suggestion: I wonder if it would be good for the dialog to have a keypress example, e.g. press the <kbd>A</kbd> key before taking the screenshot.

mcdurdin:
Note, we'll also need to copy these updates into the 18.0 developer help

Nnyny:
Done!

mcdurdin:
> Note, we'll also need to copy these updates into the 18.0 developer help

I'm sorry, this was a mistake on my part. The 18.0 help is not yet seeded... I've removed the file from 18.0 help.

PR #1289, opened 2024-05-29
chore: change url

Nnyny:


mcdurdin:
For now, let's make this change only in the 17.0 help -- as we don't yet have the 18.0 help tree available. So go ahead and remove the 18.0 file from this PR.

Nnyny:
Noted! I already removed the file.

mcdurdin (approved):

PR #1300, opened 2024-06-07
docs: remove 'Step' wording for distribution of packages

sewhite:
The wording here is unclear. "Step 1" then "Step 2" in the headings implies you need to do #1 then do #2. But these are two alternate ways to share a kmp file. You can share it directly, or via a cloud service without putting it on a website, or you can put it on a website without sharing directly, or you can do both.  My wording tries to express this.


mcdurdin (approved):

mcdurdin:
Thanks! Approving this and merging. I think that all of the packaging documentation needs significant rewriting, but sadly I am short on time to do that at present.

PR #1304, opened 2024-06-10
chore: cleanup definition lists and styling

mcdurdin:
* removes dot point so definition lists render correctly in kmc reference
* tweaks styling for code blocks in headings - background, font size

## Before
image

## After
image

### DIFF
--- a/cdn/dev/css/template.css
+++ b/cdn/dev/css/template.css
@@ -926,10 +926,6 @@ input[type="search"] {
   margin-right: 9px;
 }
 
-/*#section2 pre[class*="language-"] code[class*="language-"] {
-  line-height: 1.4;
-}*/
-
 #section2 p > code[class*="language-"],
 #section2 li > code[class*="language-"],
 #section2 dt > code[class*="language-"],
--- a/cdn/dev/css/template.css
+++ b/cdn/dev/css/template.css
@@ -924,16 +924,24 @@ input[type="search"] {
 #section2 pre[class*="language-"] {
   margin-left: 9px;
   margin-right: 9px;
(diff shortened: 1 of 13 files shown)


cdn/dev/css/template.css
```diff
@@ -924,18 +924,26 @@ input[type="search"] {
...
   margin-right: 9px;
-  padding: 9px;
-  line-height: 1.4;
 }
 
-#section2 pre[class*="language-"] code[class*="language-"] {
+/*#section2 pre[class*="language-"] code[class*="language-"] {
   line-height: 1.4;
-}
+}*/
```
darcywong00, line 931: Is this safe to remove now?

developer/17.0/reference/kmc/cli/reference.md
```diff
@@ -208,46 +208,48 @@ already be compiled.
 
 ## `kmc build windows-package-installer` additional options
 
-* `--msi <msiFilename>`
+`--msi <msiFilename>`
 
 : Full path of keymandesktop.msi to bundle into the installer. This file can be
   downloaded from https://downloads.keyman.com/windows/stable (/version).
```
darcywong00, line 214: Initially, I thought `(/version)` was a typo in a Markdown link. But now I realize it's marking a `{current-version}`.

Maybe another day, we can add `current-version` to downloads.keyman.com

mcdurdin: downloads.keyman.com is supposed to be static so I don't think we'll want to do that. If we wanted to do that, we'd add a keyman.com/go link or something like that. :grin:

Issue #94, opened 2019-12-09
Document how to transition from KMFL to Keyman for Linux


darcywong00:
From developer discussion on packaging KMFL vs Keyman for Linux with 20.04 LTS

Question raised:  Are current KMFL users expected to easily migrate to Linux Keyman, or is that not super smooth and people would be easily expected to continue using KMFL for a while?

Given there's current feedback from users that still use KMFL with their .kmn keyboards

@mcdurdin writes
> It might be good to document the transition process from KMFL to Keyman - keyboards, setup and more

Some possible steps:
1. Use kmcomp to compile .kmn to .kmx
2. Use a TBD script to generate a bare .kps package source file (XML) containing the .kmx file.

Considerations:
How to search for .kmn files (user home or system directories?)

darcywong00:
Maybe someone from the Community Site [forum](https://community.software.sil.org/c/keyman) can develop this

mcdurdin:
Do you want to make a request on the forum? I don't think this issue will be all that visible to most people otherwise.

darcywong00:
@ermshiperete provided the following on a Linux-user forum:

> However, I'd recommend switching to ibus-keyman. As you discovered that requires .kmp instead of .kmn files, but it's fairly easy to get to that:
>
>It requires installing Node (version 18.0). On Ubuntu 24.04 this can be installed with:
>
>```
>sudo apt install npm
>```
>(For older Ubuntu versions consult the [nodejs.org](http://nodejs.org/) website how to install node 18)
>
>Then install kmc, the Keyman compiler: `sudo npm install -g @keymanapp/kmc`
>
>Create a new directory for your keyboard (replace "mykeyboard" with the appropriate name...), and a source >subdirectory in that: `mkdir -p mykeyboard/source`
>
>Move your .kmn file to source/mykeyboard.kmn.
>
>Then create two files:
>
>Create a file mykeyboard.kpj:
>```xml
><KeymanDeveloperProject>
>  <Options>
>    <Version>2.0</Version>
>  </Options>
></KeymanDeveloperProject>
>```
>And source/mykeyboard.kps:
>```xml
><Package>
>  <Info>
>    <Name>MyKeyboard</Name>
>  </Info>
>  <Files>
>    <File>
>      <Name>../build/mykeyboard.kmx</Name>
>    </File>
>  </Files>
>  <Keyboards>
>    <Keyboard>
>      <Name>MyKeyboard</Name>
[...]

PR #1335, opened 2024-06-26
Add KB article on migrating from KMFL to Keyman for Linux

darcywong00:
Addresses #94 on helping Linux users migrate from KMFL to Keyman for Linux

First commit was @ermshiperete 's answer on the [community site forum](https://community.software.sil.org/t/how-to-install-an-kmn-source-kmfl-file-onto-ubuntu-24-04/8864/5) (same response on Linux-user group.

Additional commits for formatting, and tweaking the section about a kmc compilation error.

### DIFF
--- a/knowledge-base/kb0114.md
+++ b/knowledge-base/kb0114.md
@@ -1,3 +1,7 @@
+# HOWTO: Migrate from KMFL to Keyman for Linux
+
+KMFL hasn't been maintained for many years so we recommend users perform the following steps to migrate .kmn files to .kmp files and use Keyman for Linux.
+
 With a little more work this can be entirely done on Linux with Keyman 17+:
 
 ## Prerequisites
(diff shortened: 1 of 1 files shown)


ermshiperete:
See also https://github.com/keymanapp/keyman/issues/11857

ermshiperete (approved):

ermshiperete:
Just had a look at the KB page and noticed that the last paragraph ("To properly do this without an error, ...") is a bit misleading when it comes under the "Use Keyman for Linux to install the .kmp keyboard package" section header. We should either move that paragraph above the section header, or add a new section header above this paragraph.

knowledge-base/kb0114.md
```diff
@@ -0,0 +1,87 @@
...
+- Move your .kmn file to `source/mykeyboard.kmn`.
+
+- Build with:
+
+  ```bash
+  kmc build mykeyboard.kpj
+  ```
+
+Disregard the error: `"TypeError: Cannot read properties of undefined (reading 'description')"`, as the compiler will already have created the .kmp file in the `build`
+subdirectory.
```
mcdurdin, line 69: keymanapp/keyman#11856

knowledge-base/kb0114.md
```diff
@@ -0,0 +1,87 @@
...
+```bash
+km-package-install -f build/mykeyboard.kmp
+```
+
+or with the Keyman Configuration tool.
+
+To properly do this without an error, you'll have to add some additional fields to the .kps file. It's easiest to let the `kmconvert` tool generate a template which will contain all necessary the fields:
+
+```bash
+kmconvert template -id mykeyboard
```
mcdurdin, line 84: Ref keymanapp/keyman#11014

PR #1343, opened 2024-06-27

keyman-server:


keyman-status (approved):

darcywong00:
Ahh, I updated `fv_severn_obijwa_rdot` in keymanapp/keyboards#2852 but now notice I also need to handle `fv_ojibwa_rdot`

Will update...

Issue #1341, opened 2024-06-27
docs(windows): How To - Download and Install Keyman 17.0 is using version 14.0


Meng-Heng:
For reference: https://help.keyman.com/products/windows/current-version/start/download-and-install-keyman

PR #1378, opened 2024-07-03
 docs: change two images in the how-to-download-and-install-keyman

Meng-Heng:
Fixes: #1341

Can I use the images like this?
Images applies to this location: https://help.keyman.com/products/windows/current-version/start/download-and-install-keyman


mcdurdin (approved):
LGTM

PR #1479, opened 2024-08-13
chore: add all version's history api

Nnyny:
part of this PR: https://github.com/keymanapp/keyman.com/pull/462

### DIFF
--- a/_includes/includes/history_utils.php
+++ b/_includes/includes/history_utils.php
@@ -76,3 +76,73 @@ function display_history($platform, $version = '1.0'){
   // Does the magic.
   echo $Parsedown->text($contents);
 }
+
+function fetch_history($version = '1.0') {
+  $url = build_downloads_keyman_com_url("api/history/$version"); 
+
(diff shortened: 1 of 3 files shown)


_includes/includes/history_utils.php
```diff
@@ -76,3 +76,73 @@ function display_history($platform, $version = '1.0'){
...
+
+  $contents = @file_get_contents($url);
+
+  if($contents === FALSE) {
+    //header('HTTP/1.0 400 Invalid parameter');
+    $contents = "Unable to retrieve current history data for this platform.";
+  }
+
+  return $contents;
+}
```
darcywong00, line 91: Since this is very similar to `get_history()`, can we refactor `get_history()` to use this `$url` if `$platform` is not given?

Then line 109 becomes something like
```php
  $contents = get_history("", $version);
```

Nnyny: Done!

mcdurdin (approved):

PR #1499, opened 2024-08-26
docs(developer): link to kmx, kmx+ file formats

mcdurdin:


ermshiperete (approved):
LGTM

darcywong00 (approved):
lgtm

Are the `file-formats/` docs eventually going to reside in help.keyman.com?

mcdurdin:
> Are the `file-formats/` docs eventually going to reside in help.keyman.com?

I don't know. We could put it there, but perhaps would be better to publish from keyman repo in the same way as we do the app documentation (and the way I want to do Keyman Developer docs)

PR #1500, opened 2024-08-26
docs: Update API's for adjusting Android longpress delay

darcywong00:
Documents Keyman Engine for Android 18.0 API changes for keymanapp/keyman#12170 and keymanapp/keyman#12185

* getLongpressDelay
* sendKeyboardOptions
* setLongpressDelay

### DIFF
--- a/developer/engine/android/18.0/KMManager/sendOptionsToKeyboard.md
+++ b/developer/engine/android/18.0/KMManager/sendOptionsToKeyboard.md
@@ -1,0 +1,41 @@
+---
+title: KMManager.sendOptionsToKeyboard()
+---
+
+## Summary
+
+The `sendOptionsToKeyboard()` method sends options like longpress delay to the KeymanWeb keyboard.
+
+## Syntax
+
+```java
+KMManager.sendOptionsToKeyboard()
+```
+
+## Description
+Use this method to update options in the KeymanWeb keyboard.
+* Number of milliseconds to trigger a longpress gesture
+
+This method requires a keyboard to be loaded for the values to take effect.
+
+## Examples
+
+### Example: Using `sendOptionsToKeyboard()`
+
+The following code illustrates the use of `sendOptionsToKeyboard()`:
+```java
+    int currentDelayTimeMS = 250;
+
+    // Store currentDelayTimeMS
+    KMManager.setLongpressDelay(currentDelayTimeMS);
+
+    // Apply the keyboard options
+    KMManager.sendOptionsToKeyboard();
+```
+
+## History
+Keyman Engine for Android 18.0: New function.
+
+## See also
+* [getLongpressDelay](getLongpressDelay)
+* [setLongpressDelay](setLongpressDelay)
--- a/developer/engine/android/18.0/KMManager/setLongpressDelay.md
+++ b/developer/engine/android/18.0/KMManager/setLongpressDelay.md
@@ -13,7 +13,7 @@ KMManager.setLongpressDelay(int longpressDelay)
 ```
 ### Parameter
 `longpressDelay`
-: The number of milliseconds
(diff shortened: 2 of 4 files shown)


developer/engine/android/18.0/KMManager/setLongpressDelay.md
```diff
@@ -0,0 +1,42 @@
...
+Use this method to store how many milliseconds to trigger a longpress as a preference.
+This preference is stored at the Keyman app level, and is applied to all Keyman keyboards.
+
+## Examples
+
+### Example: Using `setLongpressDelay()`
+
+The following code illustrates the use of `setLongpressDelay()`:
+```java
+    int currentDelayTimeMS = 250;
```
mcdurdin, line 28: Is there a min/max range?

developer/engine/android/18.0/KMManager/sendOptionsToKeyboard.md
```diff
@@ -0,0 +1,41 @@
...
+
+```java
+KMManager.sendOptionsToKeyboard()
+```
+
+## Description
+Use this method to update options in the KeymanWeb keyboard.
+* Number of milliseconds to trigger a longpress gesture
+
+This method requires a keyboard to be loaded for the values to take effect.
```
mcdurdin, line 19: Why would this be needed? Why can't KMManager handle this itself internally?

mcdurdin (approved):

developer/engine/android/18.0/KMManager/setLongpressDelay.md
darcywong00: The Keyman for Android menu has a range of 300ms to 1500 ms. But the API doesn't have any gating on what keyboard developers may use.

developer/engine/android/18.0/KMManager/sendOptionsToKeyboard.md
darcywong00: There may be other keyboard preferences that change/get stored.

This API propagates the value to KeymanWeb to use.

developer/engine/android/18.0/KMManager/setLongpressDelay.md
mcdurdin: I was hoping this information would be added to the docs. Also, we should really be putting limits into public API functions.

developer/engine/android/18.0/KMManager/sendOptionsToKeyboard.md
mcdurdin: KMManager already knows the value though -- so why can't it be responsible for telling KMW? It seems like exposing internals to have this API at all (and I know I approved the previous PR, just thinking about this a bit more now, sorry!). I think we should review that because it's definitely less than ideal to push this responsibility of managing internal preference updates onto the app.

developer/engine/android/18.0/KMManager/setLongpressDelay.md
darcywong00: Yeah, we can document a range and add gating to KMManager. 

300-1500 ms? Or will somebody eventually want a 2000ms delay?

mcdurdin: I think we should match what Keyman app supports

developer/engine/android/18.0/KMManager/sendOptionsToKeyboard.md
darcywong00: I was waiting on merging this PR to see if we needed a Monday design discussion about the interactions between: Keyman Engine API + KeymanWeb + UI

mcdurdin: Can you write up a spec document on the various bits and pieces -- where we are at today?

PR #1513, opened 2024-08-30
chore: minor tweaks to all-versions history

mcdurdin:


_includes/includes/history_utils.php
```diff
@@ -92,44 +96,4 @@ function assignHeadingIDs($content) {
   return $content;
 }
 
-function display_version($version = '1.0'){
```
mcdurdin, line 95: I merged display_version and display_history because they were very similar

version-history/all-versions.php
```diff
@@ -1,13 +1,12 @@
 <?php
+// <h2>All platforms</h2>
```
mcdurdin, line 2: This shows 'All platforms' in the index on left-hand side of the page, a bit of a back but it works

_includes/includes/history_utils.php
```diff
@@ -70,11 +68,17 @@ function display_history($platform, $version = '1.0'){
   $contents = preg_replace('/^## /m', '### ', $contents);
   $contents = preg_replace("/\n\n\\* /", "\n* ", $contents);
 
+  // Include links to PRs
+  $contents = preg_replace('/\(#(\d+)\)/', '([#${1}](https://github.com/keymanapp/keyman/pull/${1}))', $contents);
```
mcdurdin, line 72: This is a minor tweak so we can click on the PR# in the history

darcywong00 (approved):
lgtm

PR #1518, opened 2024-09-02
chore: Revert "change: use published Keyman Engine API to fetch keyboards in keyboard help pages"

mcdurdin:
Reverts keymanapp/help.keyman.com#1272 per keymanapp/keyman#12345


darcywong00 (approved):
lgtm

PR #1563, opened 2024-09-20
fix: Skip CDN for --debug builds

darcywong00:
Mirrors keymanapp/keyman.com#471 to skip generating CDN if `./build.sh --debug`

This uses `BOOTSTRAP_VERSION v0.10` so the optional debug parameter is passed to init-container.sh

### DIFF
--- a/resources/init-container.sh
+++ b/resources/init-container.sh
@@ -5,7 +5,11 @@ rm -f keyboard/index.cache
 cd keyboard
 php -d include_path=/var/www/html/_includes:. _build_cache.php
 
-echo "---- Generating CDN ---"
-cd ../cdn
-php -d include_path=/var/www/html/_includes:. cdnrefresh.php
-cd ..
+if [[ ! $1 =~ "debug" ]]; then
+  echo "---- Generating CDN ---"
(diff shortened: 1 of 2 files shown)


resources/init-container.sh
```diff
@@ -5,7 +5,11 @@ rm -f keyboard/index.cache
...
-cd ../cdn
-php -d include_path=/var/www/html/_includes:. cdnrefresh.php
-cd ..
+if [[ ! $1 =~ "debug" ]]; then
+  echo "---- Generating CDN ---"
+  cd ../cdn
+  php -d include_path=/var/www/html/_includes:. cdnrefresh.php
+  cd ..
+else
+  echo "Skip Generating CDN"
```
mcdurdin, line 14: I just realized that we should remove cdn/deploy if it exists (e.g. from a previous run). I think:

```suggestion
  echo "Skip Generating CDN and clean CDN cache"
  rm -rf ../cdn/deploy
```

mcdurdin (changes requested):

mcdurdin (approved):

PR #1574, opened 2024-09-25
Update PrimerPrep URLs

DavidLRowe:
Request from Jeff Heath (author of PrimerPrep):
```
I noticed on https://help.keyman.com/developer/current-version/guides/lexical-models/tutorial/step-3 
that it points to lingtransoft.info for the PrimerPrep installation. 
That link should just point to https://software.sil.org/primerprep/ and/or 
https://software.sil.org/primerprep/downloads/.

Also the next line can be toned down a bit, maybe:
Run PrimerPrep (note that the first run might be slow; subsequent starts are faster)

```


DavidLRowe:
@jeffheath-sil Does this look okay?

mcdurdin (approved):

PR #1575, opened 2024-09-26
chore: convert docs dev/engi/web/17/guide

Meng-Heng:
Relates to: https://github.com/keymanapp/keyman/issues/12347

Is it a requirement for everything to be in markdown? I could not get the right layout of the table in `developer/engine/web/17.0/guide/user-interface-design.md` that is the reason I have not change it to md.


mcdurdin:
> Is it a requirement for everything to be in markdown? I could not get the right layout of the table in `developer/engine/web/17.0/guide/user-interface-design.md` that is the reason I have not change it to md.

No, it is fine to leave complex tables as inline HTML in the Markdown.

developer/engine/web/17.0/guide/adding-keyboards.md
```diff
@@ -0,0 +1,59 @@
+---
+title: Adding Keyboards
```
mcdurdin, line 2: I note the title is different, but actually better than original :grin:

mcdurdin (approved):

Issue #1548, opened 2024-09-14
bug: keyboard elements appear above the banner


mcdurdin:
![image](https://github.com/user-attachments/assets/81d087ba-a96b-46ab-8a17-511d3686bd56)

e.g. https://help.keyman.com/keyboard/beria_erfe_phonetic/1.0/beria_erfe_phonetic

PR #1608, opened 2024-10-11
fix: override KeymanWeb CSS for z-index for documentation OSK

mcdurdin:
A more comprehensive fix would need to be made to KeymanWeb, but that should be part of a more systematic review of KeymanWeb's CSS in the future.

Fixes: #1548


darcywong00 (approved):
lgtm

PR #1609, opened 2024-10-11
docs: more conversion on  developer/engine/android/18.0/KMManager

Nnyny:
partial fix of: https://github.com/keymanapp/keyman/issues/12347


darcywong00 (approved):
lgtm

PR #1684, opened 2024-11-07
chore: more /developer/core/desktop to /developer/core

mcdurdin:
Also removes /developer/core/web (but keeps redirect)


darcywong00:
> Also removes /developer/core/web (but keeps redirect)

So /developer/core/web redirects to /developer/engine/web?

darcywong00 (approved):
lgtm

mcdurdin:
> > Also removes /developer/core/web (but keeps redirect)
> 
> So /developer/core/web redirects to /developer/engine/web?

Yep

Issue #1712, opened 2024-11-25
feat: change keyman videos from link to video showcasing


Meng-Heng:
1. Look through all the web pages for Video links.
2. Change it to Display the video for 'Press Play'.
3. Choose the right format for all devices.

PR #1713, opened 2024-11-25
feat: two samples of integrating keyman videos onto help.keyman.com

Meng-Heng:
follow: #1712

video sample:

https://github.com/user-attachments/assets/77c48f4e-d7e2-4e56-b4dd-74f1578f2903

- [x] Should we use the `<iframe>` tag or use js onclick to set the `<iframe>`?

-- updated on 09-Dec-2024
Fixes: #1712


cdn/dev/js/videoloader.js
```diff
@@ -0,0 +1,7 @@
+window.addEventListener('DOMContentLoaded', (e) => {
+    let t = document.getElementById("keymanThumbnails");
+    t.addEventListener("click", function() {
```
mcdurdin, line 3: I think we need to use [`querySelectorAll`](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll) instead of `getElementById`, and match on any element that has the appropriate `data-video` attribute, so we can have multiple thumbnails in the same document. That means then we need to add an event listener for each element found, in a loop.

mcdurdin (changes requested):
I like the idea of not embedding the video until a img preview is clicked; we'll need to tweak the code slightly to get it to work across multiple thumbnails 

(This may not be relevant with the requested change, but as a side note, the `getElementById()` call result was not checked -- if a page did not have a matching element, it would return null, causing an error on `t.addEventListener` line.)

cdn/dev/js/videoloader.js
Meng-Heng: Yes, I'll look into this.

developer/videos/index.md
```diff
@@ -10,13 +10,20 @@ These videos give a walkthrough on some improvements to Keyman Developer.
 
 * [Character Map + Touch Layout Editor integration improvements](https://www.youtube.com/watch?v=qoLpuah72kw) (2:15)
 
+    <img id="keymanThumbnails" data-video="https://www.youtube.com/embed/qoLpuah72kw" src="https://img.youtube.com/vi/qoLpuah72kw/maxresdefault.jpg" width="600px">
```
mcdurdin, line 13: * `id` should not be repeated across multiple elements -- perhaps use `class` instead.
* Instead of `keymanThumbnails` can we use `video-thumbnail`?
* 
```suggestion
    <img class="video-thumbnail" data-video="https://www.youtube.com/embed/qoLpuah72kw" src="https://img.youtube.com/vi/qoLpuah72kw/maxresdefault.jpg" width="600px">
```

cdn/dev/js/videoloader.js
```diff
@@ -0,0 +1,13 @@
...
+  thumbnailsList.forEach(each => {
+  	each.addEventListener('click', () => {
+
+  		if(each.hasAttribute('data-video')) {
+        let videoId = each.getAttribute('data-video')
+    		each.outerHTML = "<iframe width='600' height='315' src='" + videoId + "' allowfullscreen></iframe>";
+      }
+    })
+  })
+})
```
mcdurdin, line 13: Use an [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) to prevent pollution of global namespace:
```suggestion
(function() {
  const thumbnailsList = document.querySelectorAll(".video-thumbnail");
  
  window.addEventListener('DOMContentLoaded', () => {
    thumbnailsList.forEach(thumbnail => {
  	  thumbnail.addEventListener('click', () => {
  
  		  if(thumbnail.hasAttribute('data-video')) {
          let videoId = thumbnail.getAttribute('data-video')
    		  thumbnail.outerHTML = "<iframe width='600' height='315' src='" + videoId + "' allowfullscreen></iframe>";
        }
      })
    })
  })
})();
```

* I've also renamed the variable `each` to `thumbnail` to better represent what it is.
* I renamed the selector from `#keymanThumbnails` to `.video-thumbnail` (a class rather than an id, and singular rather than plural, and 'video' instead of 'keyman' to clarify what it is for, and finally, using kebab-case instead of camelCase :grin:)

mcdurdin (commented):
This is looking good. A few minor style changes requested!

Meng-Heng:
Thanks @mcdurdin. I'm taking notes, this is so unfamiliar to me. 😅

mcdurdin (approved):
LGTM!

PR #1719, opened 2024-11-26
chore: add new issues to Keyman project automatically

mcdurdin:
Fixes: #1720
Relates-to: keymanapp/keyman.com#505


mcdurdin:
Tested this after merge and it worked correctly, so will now replicate across other repos in the same way.

PR #1744, opened 2024-12-07
feat: move to GFM-style alerts

mcdurdin:
* v0.12 of shared-sites
* adds CSS for GFM-style alerts to render nicely


darcywong00 (approved):
lgtm

So each site will have to add template.css to get the GM-style alerts?

mcdurdin:
> So each site will have to add template.css to get the GM-style alerts?

For now, yes. We don't have shared CSS at this point -- we need to be _very_ careful sharing CSS because the templates are overloaded and getting fragile. So given the alerts are mostly useful for documentation, I think it's fine to make them available only on help.keyman.com at this time.

PR #1755, opened 2024-12-16
fix: links to the available version of docs

Meng-Heng:
Solving: https://github.com/keymanapp/keyman.com/issues/415.

This PR is ready for review.


developer/language/guide/virtual-keys.md
```diff
@@ -167,10 +167,9 @@ The presence of this store also ensures that Caps Lock is handled correctly for
 KeymanWeb keyboards (see
 [#5460](https://github.com/keymanapp/keyman/issues/5460#issuecomment-966602098)).
 
-## Virtual character keys {#virtual-character-keys}
+## Virtual character keys
```
darcywong00, line 170: We might need to update all the links to these anchors. This one corresponds to l.84

which might work if we change that to `(#toc-virtual-character-keys).

I'm not sure we want to change all the links this way though.

darcywong00 (commented):
Unsure how to handle anchored links

developer/language/guide/virtual-keys.md
Meng-Heng: Hello @darcywong00. The changes of this PR shows this:
image

Do you mean adding `{#toc-` in front of `virtual-character-keys}`, not removing them entirely? Also, may I know what is I.84? 😅

darcywong00: The abbreviation l.84 means "line 84" in the file (above). 
```md
keyboard, along with the specified modifier. [More info](#virtual-character-keys)
```

When I tried locally, clicking the "More info" links didn't work

I noticed the links in "On this page" did work though. And they had the toc- prefix...

Meng-Heng: Thank you @darcywong00. I've seen the abbreviation once before but couldn't remember it. 

Would this fix be appropriate? the {#virtual-character-keys} showed up like this if not removed.
image

I am trying {#virtual-character-keys} on "More info" but no success.
image

Only being scrolled to via clicking on "On this page", but the URL shows this:
`http://help.keyman.com.localhost:8055/developer/language/guide/virtual-keys#toc-virtual-character-keys-virtual-character-keys-`

I think `#toc-virtual-character-keys` was assigned as an `id` before converting to markdown.

mcdurdin: > `#toc-virtual-character-keys`

The `#toc-` prefix is generated automatically and used for navigation internally in the page. The other hashtags were intended to be used as permanent anchors but it looks like they are not working.

mcdurdin: We need to restore these anchors in this PR. See #1791.

mcdurdin (changes requested):

mcdurdin (approved):

PR #1771, opened 2025-01-13
chore: broken internal URLs

Meng-Heng:
Continue: https://github.com/keymanapp/keyman.com/issues/415

I'm using an audit result to determined the broken internal URLs. Please let me know if this is not a correct URL. Thanks!
image


darcywong00:
Maybe there's a few more though?
https://github.com/keymanapp/help.keyman.com/blob/master/developer/15.0/context/kmcomp.php

I didn't check how far back

Meng-Heng:
Thanks @darcywong00,

It is way back to 11.0. I am putting them separately #1777.

PR #1821, opened 2025-01-27
fix: prevent KeymanWeb load of keyboards that do not have .js versions

mcdurdin:
Also adds a console.error catch() reporter so we don't get unhandled promise rejections in the event that the keyboard fetch fails for any reason.

Fixes: keymanapp/keyman#13043

### DIFF
--- a/cdn/dev/js/kbd-docs.js
+++ b/cdn/dev/js/kbd-docs.js
@@ -70,11 +70,8 @@ function loaded(){
     link.css('margin-top',diff);
   }
 
-  (function(){
+  function loadKeymanWebKeyboard(keyboardName) {
     // Insert keyboard into osk div
-
-    var url = location.pathname.split('/');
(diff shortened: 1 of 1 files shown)


cdn/dev/js/kbd-docs.js
```diff
@@ -175,12 +178,17 @@ function loaded(){
...
+    // not just in Web's OSK.
     // TODO: consider using the retrieved stub to get that in future
     $.ajax({
       url: 'https://api.keyman.com/keyboard/' + keyboardName,
       success:function(data) {
         if(typeof data != 'object') return;
+
+        if(data.platformSupport?.desktopWeb || data.platformSupport?.mobileWeb) {
+          loadKeymanWebKeyboard(keyboardName);
+        }
```
mcdurdin, line 190: This is where the test for a keyboard supporting web has been added, so this call now happens before the KeymanWeb call.

jahorton (approved):

PR #1846, opened 2025-02-07
chore: add version 17 details to language history

mcdurdin:


developer/language/guide/history.md
```diff
@@ -59,43 +65,71 @@ title: Keyman language history
...
+
+*   Initial support for LDML keyboard format in compiled files. Note that this
+    is not a change to the Keyman keyboard language.
+
+## Keyman 17
+
+*   Support for [flick and multi-tap gestures](/developer/current-version/reference/file-types)
+    in touch layouts
+*   Support for additional special key caps in
+    [touch layouts](/developer/current-version/reference/file-types/keyman-touch-layout):
```
darcywong00, line 132: 1. Could we include the "Key text" anchor? (the Markdown that corresponds to `keyman-touch-layout#toc-key-text`)

2. I understand that `current-version` is more flexible to apply to future versions. If we pin to 17.0, would that avoid future links breaking? (hypothetically a future Keyman version renames keyman-touch-layout to something else)

developer/language/guide/history.md
```diff
@@ -59,43 +65,71 @@ title: Keyman language history
...
+    [`&oldLayer` system store](../reference/oldlayer)
+
+## Keyman 16
+
+*   Initial support for LDML keyboard format in compiled files. Note that this
+    is not a change to the Keyman keyboard language.
+
+## Keyman 17
+
+*   Support for [flick and multi-tap gestures](/developer/current-version/reference/file-types)
```
darcywong00, line 129: Could you help clarify why gestures links to file-types?

Maybe we add a `gestures` section after `subkey` in keyman-touch-layout?

developer/language/guide/history.md
mcdurdin: 1. OK
2. We could, but then we would have to go and update with each release, which is a pain. With current-version, at least the link checker will complain if we move things around.

developer/language/guide/history.md
mcdurdin: It's the only reference information for gestures I could find. Also not great, so have updated to the context help for the editor for now.

mcdurdin:
@darcywong00 have addressed review comments

darcywong00 (approved):
lgtm

Issue #1870, opened 2025-02-13
refactor: stop using `$TestServer` and related variables


mcdurdin:
There are only a few references left to `$TestServer`, and there are a few other very old related variables that are probably unused. It would be good to get rid of them as a part of the work done in KeymanHosts (e.g. keymanapp/shared-sites#53) to make future maintenance of this simpler.

This issue covers help.keyman.com and keymanweb.com.

```
12 results - 4 files

help.keyman.com\_includes\includes\__kmwheader.php:
  83      default:
  84:       global $TestServer;
  85:       if($TestServer) {
  86          return false;   // We should never get this except briefly while documenting a new major version.

help.keyman.com\_includes\includes\servervars.php:
  30  
  31:   $TestServer = false;
  32  

  37    function cdn($file) {
  38:     global $cdn, $TestServer;
  39:     $use_cdn = !$TestServer || (isset($_REQUEST['cdn']) && $_REQUEST['cdn'] == 'force');
  40      if($use_cdn && $cdn && array_key_exists('/'.$file, $cdn)) {

keymanweb.com\inc\servervars.php:
  38      // $site_protocol is used only by util.php at this time.
  39:   $TestServer = (KeymanHosts::Instance()->Tier() == KeymanHosts::TIER_DEVELOPMENT) ||
  40      (KeymanHosts::Instance()->Tier() == KeymanHosts::TIER_TEST) ? true : false;
  41:   $site_protocol = $TestServer ? 'http://' : 'https://';
  42  

  52    function cdn($file) {
  53:     global $cdn, $staticDomain, $TestServer;
  54:     $use_cdn = !$TestServer || (isset($_REQUEST['cdn']) && $_REQUEST['cdn'] == 'force');
  55      if($use_cdn && $cdn && array_key_exists('/'.$file, $cdn)) {

keymanweb.com\inc\util.php:
  33      }
  34:     global $TestServer;
  35:     if(isset($TestServer) && $TestServer) { echo "<pre>"; debug_print_backtrace();echo "</pre>"; }
  36      exit('Invalid parameter');
```

Other variables to eliminate in keymanweb.com are:

```
  $site_suffix 
  $site_protocol 
  $url_keymanweb_res 
  $staticDomainRoot
  $site_keymanwebhelp
  $site_keymanwebdemo
  $site_keyman       
  $staticDomain
```

Other variables to eliminate in help.keyman.com are:

```
  $keyman_com
  $site_protocol 
[...]

PR #1894, opened 2025-02-24
refactor: remove use of old $TestServer variable

mcdurdin:
Also removed $site_securetavultesoft and $downloadsDomain.

Fixes: #1870

### DIFF
--- a/_includes/includes/__kmwheader.php
+++ b/_includes/includes/__kmwheader.php
@@ -72,6 +72,7 @@ function get_kmwbuild()
     return $kmwbuild;
   }
 
+  // TODO: rewrite this -- it's kinda outdated
   // In case of no match, the defaults:
   switch($maj_version) {
     case 11:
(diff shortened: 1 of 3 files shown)


_includes/includes/__kmwheader.php
```diff
@@ -81,8 +82,7 @@ function get_kmwbuild()
     case 2:
       return "2.0.473";
     default:
-      global $TestServer;
-      if($TestServer) {
+      if(KeymanHosts::Instance()->Tier() != KeymanHosts::TIER_PRODUCTION) {
```
darcywong00, line 85: I only saw `$TestServer` get initialized to false in servervars.php.

Would we ever check TIER_TEST?

mcdurdin: `$TestServer` was closer to staging -- it derives from when we had a 'test server' hosting tavultesoft's websites years and years ago. It wasn't used for automated tests. I don't think we need to check TIER_TEST here -- this code needs refactoring and rewrite one day because it's very old and the values are really out of date!

darcywong00 (approved):
lgtm

PR #1909, opened 2025-03-05
Small clarifications to phphelpfile.md

alex-larkin:
I have suggested some small clarifications, based on my experience of using this tutorial for PR keymanapp#3354. In particular, giving a fuller list of what HTML tags to delete and how to embed the CSS inside the `<?PHP ?>` tag. (I also split step 5 into two steps, and hence following steps have been renumbered.)


darcywong00 (approved):
lgtm

Thanks for the contribution!

PR #1911, opened 2025-03-08
Slight tweak to phphelpfile.md

alex-larkin:
My previous revision instructed to delete `<html>` tag and its closing tag along with all of its contents. That technically means deleting the entire document. This rewording corrects that.


darcywong00 (approved):
lgtm

PR #2013, opened 2025-04-23
fix: Disallow all bots since staging site doesn't need to be indexed

darcywong00:
Staging site getting hammered by an Amazon bot causing an OOM bootloop

@mcdurdin recommends we disallow all bots since the staging site doesn't need to be indexed.


mcdurdin (approved):

mcdurdin:
We should consider this on other staging sites as well

PR #2021, opened 2025-04-25
chore: Update fallback version of KeymanWeb to 18.0.235

darcywong00:
In the Development tier, if the other ancillary Keyman sites aren't running, then the site will fetch this fallback version of KeymanWeb from s.keyman.com

The previous version '13.0.108' doesn't render the keyboard layouts

### DIFF
--- a/_includes/includes/head.php
+++ b/_includes/includes/head.php
@@ -82,7 +82,7 @@
    else window.onload = downloadJSAtOnload;
   </script>
   <?php if(isset($IncludeKeymanWeb) && $IncludeKeymanWeb == true){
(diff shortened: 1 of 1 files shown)


mcdurdin (approved):

darcywong00:
> ~In the Development tier, if the other ancillary Keyman sites aren't running,~ then the site will fetch this fallback version of KeymanWeb from s.keyman.com

ah, to clarify, I've been modifying my local KeymanHosts.php file to use the live s.keyman and api.keyman sites.
The fallback version is still what gets used.

PR #2069, opened 2025-05-16
feat: add top level page for LDML keyboard and first few pages ✍️ 

srl295:
- adds /developer/ldml/guide and /developer/ldml/reference
- add a new top level link from /developer
- bulk out /developer/ldml/reference with every element (mostly stubs)
- add docs for keyboard3

For #13836 

@keymanapp-test-bot skip


mcdurdin:
This is a good start. I wonder if this could be the seed for an epic branch? I would prefer not to merge the docs until they have some real content, but there is a benefit in having a review of each step in the process. Will consider and discuss with you.

srl295:
> This is a good start. 

Sorry, this was meant to be draft - fixed.

> I wonder if this could be the seed for an epic branch? I would prefer not to merge the docs until they have some real content, but there is a benefit in having a review of each step in the process. Will consider and discuss with you.

right, probably a good idea.

srl295:
OK, i think this should be good to merge to the epic?

PR #2074, opened 2025-05-19
fix: remove documentation for never implemented feature of `if`

ermshiperete:
While discussing [#13492](https://github.com/keymanapp/keyman/pull/13492) we discovered that the comparison with a store was never implemented. This change fixes the documentation to match reality.

Test-bot: skip


darcywong00 (approved):
lgtm

PR #2079, opened 2025-05-21
chore: update BOOTSTRAP_VERSION to v0.28

keyman-server:
This PR updates the BOOTSTRAP_VERSION in build.sh to v0.28.


mcdurdin:
Will be replaced, this was to verify a release workflow (see keymanapp/shared-sites#64)

Issue #2080, opened 2025-05-21
bug: Edit Page button edits help.keyman.com page instead of keyman page for app docs


mcdurdin:
- at least for developer (see #2078 for example)
- we need to update the edit script to point to the appropriate place in the keyman repo for editing product documentation
- don't forget engine docs also

ermshiperete:
See also #1999

mcdurdin:
> See also [#1999](https://github.com/keymanapp/help.keyman.com/issues/1999)

That's what I was looking for before, and missed!

PR #2093, opened 2025-05-23
fix: edit correct page for developer, engine help

mcdurdin:
Also notes difficulties for editing keyboards and alternate way of storing metadata to make determining path to edit easier to find.

Fixes: #2080

### DIFF
--- a/_includes/includes/com.keyman.help.editpage.inc.php
+++ b/_includes/includes/com.keyman.help.editpage.inc.php
@@ -11,6 +11,7 @@
     $base = "https://github.com/keymanapp/help.keyman.com/edit/master";
     $app =  "https://github.com/keymanapp/keyman/edit/master";
 
+    // Map from help.keyman.com URL to path in keymanapp/keyman repository
     $map = [
       'products/android' => 'android/docs/help/',
       'products/iphone-and-ipad' => 'ios/docs/help/',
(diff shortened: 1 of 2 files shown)


_includes/includes/com.keyman.help.editpage.inc.php
```diff
@@ -1,69 +1,59 @@
...
+
+  // TODO: it may be better to have a file in the root of each folder which
+  //       determines the root path for editing and any parameters relating
+  //       to it.
+
   function getPageUrlForEditLink($self, $id, $file) {
     $base = "https://github.com/keymanapp/help.keyman.com/edit/master";
     $app =  "https://github.com/keymanapp/keyman/edit/master";
 
     $map = [
```
ermshiperete, line 14: This could use a comment. I guess it maps from the path in help.keyman.com to the path in keyman with the version number determining the branch?

_includes/includes/com.keyman.help.editpage.inc.php
```diff
@@ -1,69 +1,59 @@
...
     }
-    else if(preg_match('/^products\/('.$maps.')\/(\d+\.\d+)\/(.*)\.md$/', $file, $matches) && version_compare($matches[2], '14.0', '>=')) {
-      // Product help, version 14.0 and later
+    else if(preg_match('/^('.$maps.')\/(\d+\.\d+)\/(.*)\.md$/', $file, $matches) && version_compare($matches[2], KeymanVersion::stable_version, '>=')) {
+      // Product, Developer, Engine help, version 14.0 and later. Note that this
+      // is not a perfect match and in some cases will have missing links
+      // because we are not attempting to check for latest version of help
       return "$app/{$map[$matches[1]]}{$matches[3]}.md";
     }
+    else if(preg_match('/^(products\/|developer\/engine\/.+|(developer\/(\d+\.\d+)))/', $file, $matches)) {
```
ermshiperete, line 39: Can't we use the same pattern as above?

```suggestion
    else if(preg_match'/^('.$maps.')\/(\d+\.\d+)\/(.*)\.md$/', $file, $matches)) {
```

mcdurdin: This is a broader match -- also captures index pages, etc. Will add a comment.

PR #2119, opened 2025-06-12
chore: additional deadkey example from community site

Meng-Heng:
I don't want the example from https://community.software.sil.org/t/how-can-we-set-branch-transformation-in-keyman/9771/4 get lost.

This PR is ready for review.

### DIFF
--- a/developer/language/reference/deadkey.md
+++ b/developer/language/reference/deadkey.md
@@ -67,9 +67,23 @@ dk()* any(char) dk()* + [K_BKSP] > nul
 
 ### Example: Using `deadkey()`
 
-```
+```c
 + '`' > dk(backquote)
 dk(backquote) + 'e' > 'è'
+```
+
+### Example: Using `dk()`
+An example for turning the two sequences (w and ew) into the same character, but distinguish them so that you can treat them differently in future matches.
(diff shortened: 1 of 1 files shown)


developer/language/reference/deadkey.md
```diff
@@ -67,11 +67,25 @@ dk()* any(char) dk()* + [K_BKSP] > nul
...
 
-```
+```c
 + '`' > dk(backquote)
 dk(backquote) + 'e' > 'è'
 ```
 
+### Example: Using `dk()`
+An example for turning the two sequences (w and ew) into the same character, but distinguish them so that you can treat them differently in future matches.
+```c
```
ermshiperete, line 77: Why use the C language identifier for this code block? I think it would be better to leave that out:

````suggestion
```
````

File: developer/language/reference/deadkey.md [hunk 2]
```diff
@@ -67,11 +67,25 @@ dk()* any(char) dk()* + [K_BKSP] > nul
 
 ### Example: Using `deadkey()`
 
-```
+```c
```
ermshiperete, line 70: See below
````suggestion
```
````

PR #2120, opened 2025-06-12
chore: add download shared font folder and use alerts

Meng-Heng:
Inspired from https://community.software.sil.org/t/how-to-update-keyboard-without-cloning-the-entire-repository/9793/2

This PR is ready for review.

### DIFF
--- a/knowledge-base/kb0106.md
+++ b/knowledge-base/kb0106.md
@@ -24,5 +24,10 @@ your computer.
 > [!Note]
 > Also that attempting to download a large keyboard (or group of keyboards) may fail due to the large size.
 
+## Folder Structure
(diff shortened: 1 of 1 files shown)


LornaSIL:
Looks good! You might want to add a bit more about the folder structure? Maybe put it just above "See Also"?

> ## Folder Structure
>When downloading both the keyboard and the fonts, you will want to duplicate the normal repo structure. For example, in the `sil_ipa` example, you would have a folder structure of `release/sil/sil_ipa` where you would put the contents of the keyboard zip file. Then you would need a folder called `release/shared/fonts/sil/charis` where you would put the contents of the font zip file.

>Following this folder structure will allow your local build to work the same as in the Keyman build system.

LornaSIL (approved):
LGTM (Thanks!)

PR #2123, opened 2025-06-16
chore: add tips and caution of flicks and long presses docs

Meng-Heng:
Information from https://community.software.sil.org/t/longpress-flicks-one-or-the-other-or-both/10065/3

This PR is ready for review.


Meng-Heng:
My plan was to update both locations. I'll see where I can put this and make a pull request to only the keyman repo.
Thank you.

PR #2135, opened 2025-06-24

keyman-server:


mcdurdin:
CI failure fixed in https://github.com/keymanapp/keyboards/pull/3592, will be addressed in follow-up PR.

PR #2234, opened 2025-09-01
maint: enable automatic version links for renamed files

ermshiperete:
This change adds a mapping between old and new filenames. If a file doesn't exist for a particular version we also try with the mapped filename.

Related: https://github.com/keymanapp/keyman/pull/14646
Test-bot: skip

### DIFF
--- a/_includes/includes/page-version-file-mappings.php
+++ b/_includes/includes/page-version-file-mappings.php
@@ -1,5 +1,7 @@
 <?php
-  // Mappings for files that got renamed at some point
+  // Mappings for files that got renamed at some point. These mappings are applied
+  // for the version switching mechanism on pages such as /developer/x.y/.... They
+  // are applied to _all_ versioned folders, so be careful when adding new mappings.
   global $PageVersionFileMappings;
   $PageVersionFileMappings = [
     ['whats-new', 'whatsnew'],


_includes/includes/page-version.php
```diff
@@ -1,5 +1,13 @@
 <?php
 
+  // Mappings for files that got renamed at some point. Has to be in
+  // both directions.
+  global $FileMappings;
+  $FileMappings = [
+    'whats-new' => 'whatsnew',
+    'whatsnew' => 'whats-new',
+  ];
```
mcdurdin, line 9: This could go in a separate file. Also given it is a global variable, can we call it `$PageVersionFileMappings`?

And ... I wonder if we could make it easier to maintain -- and support multi-way mappings, by having arrays of mappings:

```suggestion
  // Mappings for files that got renamed at some point. Has to be in
  // both directions.
  global $FileMappings;
  $PageVersionFileMappings = [
    ['whats-new', 'whatsnew'],
  ];
```

mcdurdin (changes requested):
I think this is a reasonable hammer for solving our problem.

* It is disconnected from the file content so runs the risk of becoming hard to maintain in the future as renames get more complex (see [.htaccess](https://github.com/keymanapp/help.keyman.com/blob/75f42230886570ac952b6acf2b5bdb21cc45df07/.htaccess) for concrete example of the pain).

* I am a little concerned that in many cases the rename is a one-time version change; see for example https://github.com/keymanapp/help.keyman.com/blob/814d5fd36eb6769bb4dabe10c7882e01eb515af7/products/windows/.htaccess#L5-L13. 

* Also, this is site-wide, but allows for no folder-specific mappings.

* Requesting that the `$FileMappings` array become a multi-dimensional array of synonyms.

ermshiperete:
Yes, as it is it doesn't work for more complex mappings or mappings that contains paths. I decided to practice YAGNI and not implement it until we need it.

PR #2263, opened 2025-09-16
docs: layouts for hardware forms of keyboards

mcdurdin:
Test-bot: skip


darcywong00:
hmm, the enter key doesn't seem to get styled on my locale site

image

mcdurdin:
> hmm, the enter key doesn't seem to get styled on my locale site

Good catch ... worked around that - was a Parsedown parse bug I think. I should have fired up my local site for testing and not just relied on VS Code presentation...

Meng-Heng:
Would it be good to get the key IDs to fit onto the keycap? I'm getting an overflowing behavior. Enter key is good now.
Screenshot 2568-09-17 at 10 49 52 in the morning

darcywong00 (approved):
lgtm

mcdurdin:
@Meng-Heng where are you getting that screenshot from? help.keyman.com.localhost, or something else? I don't get that font on my tests

Meng-Heng:
I got it from help.keyman.com.localhost, @mcdurdin.

PR #2311, opened 2025-10-28
docs: context + keystroke model (vs sequences)

Meng-Heng:
Requested by @mcdurdin,

> https://community.software.sil.org/t/how-to-code-in-ligatures-in-keyman/10905
> This question keeps coming up! Would you like to add a page to the documentation on the context + keystroke model (vs key sequences)? We have https://help.keyman.com/developer/language/guide/rules, but it might also be helpful to add an example or two at the top of that showing how a sequence could be implemented, something like:
> ```
> + 'A' > 'Ж'
> + 'B' > 'Z'
> 'Z' + 'A' > 'տ'
> ```

Test-bot: skip

### DIFF
--- a/developer/language/guide/rules.md
+++ b/developer/language/guide/rules.md
@@ -47,9 +47,9 @@ example, in a Dinka keyboard, the following rule transforms "ee" on screen, when
 ### A sequence of keystrokes
 
 It is common to think of keyboard rules in terms of keystroke sequences. For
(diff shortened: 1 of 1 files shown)


mcdurdin:
Thanks @Meng-Heng for starting this documentation update. I have added some additional examples. 

All reviewers: do these examples make sense? do you think they'll help people starting off with keyboard rules?

mcdurdin:
I went with Armenian, even though two of the output characters are similar, they are not identical. Perhaps this will cause readers to check more carefully?

PR #2344, opened 2025-11-28
chore: Update build.sh test

darcywong00:
Mirrors keymanapp/keyman.com#578 changes and relates to keymanapp/keyman.com#493
Also :cherries: picks keymanapp/keyman.com#624 to emit broken-link-checker logs before summarizing

> 2. Make it possible to lint and phpunit test from local dev machine
> 
> This should also be done on help.keyman.com, elsewhere as well?

composer.json was already clean, but this PR updates `build.sh test` to do:
* PHP unit tests
* Lint
* Link checking

This involves passing `build.sh build --debug` parameter so the `composer install` step installs dev dependencies.

Test-bot: skip


darcywong00:
hmm - troubleshooting why the GHA (tier.TEST) is using cdn/deploy instead of cdn/dev

PR #2347, opened 2025-12-01
docs: Use /cdn/dev assets in older iOS engine guides

darcywong00:
Relates to keymanapp/keyman#15253

The `/cdn/deploy` assets don't exist when the help.keyman site is built with `./build.sh build --debug`

This updates older Keyman Engine for iPhone and iPad guides to use `/cdn/dev/` asset links since they'll always exist.
19.0 content handled by keymanapp/keyman#15253

Test-bot: skip


Meng-Heng (approved):
LGTM.

PR #2392, opened 2026-01-14
feat: add search knowledge base title

Meng-Heng:
I made this to be able to search for the kb article's title.

This PR is ready for reviews.

Test-bot: skip

Before search
Screenshot 2569-01-14 at 12 09 24 in the afternoon

When search
Screenshot 2569-01-14 at 12 09 39 in the afternoon

### DIFF
--- a/knowledge-base/index.php
+++ b/knowledge-base/index.php
@@ -71,13 +71,13 @@
         $handle = fopen($kb, "r");
         $title = fgets($handle);
         $title = substr($title, 2);
+        fclose($handle);
 
         $searchTarget = strtolower("KMKB{$id} {$title}");
         if ($query !== '' && stripos($searchTarget, $query) === false) {
           continue; // skip this if there isn't a query and a query match
         }
 
-        fclose($handle);
         echo "<li><a href='".link_from_id($id)."'>KMKB{$matches[1]}: " . htmlspecialchars($title) . "</a></li>";
       }
     }


Meng-Heng:
@ermshiperete, I decided to add the query onto the input instead. Would this be fine?
Screenshot 2569-01-15 at 2 14 25 in the afternoon

knowledge-base/index.php
```diff
@@ -55,19 +55,46 @@
...
     foreach($kbs as $kb) {
       if(preg_match("/^kb(\d+)\.md$/", $kb, $matches)) {
         $id = $matches[1];
         $handle = fopen($kb, "r");
         $title = fgets($handle);
         $title = substr($title, 2);
+
+        $searchTarget = strtolower("KMKB{$id} {$title}");
+        if ($query !== '' && stripos($searchTarget, $query) === false) {
+          continue; // skip this if there isn't a query and a query match
```
ermshiperete: Do we have to `fclose($handle);` before we `continue` so that we don't leak $handle?

ermshiperete: In the worst case the server uses more and more resources and eventually can crash because it runs out of file handles. Whether or not that happens depends on the implementation of the server, how often processes get restarted etc, but leaks always smell.

I'm not that familiar with php and don't know if `fclose` is mainly cosmetic and php does some kind of garbage collection in the background, but if there is a `fclose` and we call it elsewhere, I'd call it here as well.

Meng-Heng: Thank you for the explanation

PR #2502, opened 2026-05-08
docs(linux): Update KB article about using Keyman with KDE

ermshiperete:


mcdurdin (approved):

rc-swag (approved):
lgtm 
But did not actually test steps on linux box

PR #2516, opened 2026-05-21
docs: improve documentation on `store()`

### LINKED ISSUE #2374: docs: improve documentation on `store()`
https://help.keyman.com/developer/language/reference/store

Currently we don't explain that `value` is used to initialize the store (that's mentioned under `save()`, but not here). Also it would be good to explain that any persisted value stored previously with `save()` will take precedence over `value`.

ermshiperete:
Fixes: #2374
Test-bot: skip

### DIFF
--- a/developer/language/reference/store.md
+++ b/developer/language/reference/store.md
@@ -18,9 +18,13 @@ store(storeName) value
 :   The name of the store to be defined.
 
 `value`
-:   An extended string value. This will be the initial default value.
-    Any persisted value stored previously with [`save()`](save) will
-    take precedence over `value`.
+:   An extended string value.
+
+    If used to declare a [variable store](../guide/variable-stores), this 
+    will be the initial default value. Any persisted value stored 
+    previously with [`save()`](save) will take precedence over `value`. 
+    Note that variable stores should not be used with [`any()`](any), 
+    [`index()`](index), or [`outs()`](outs).
 
 ## Description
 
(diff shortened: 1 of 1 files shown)


developer/language/reference/store.md
```diff
@@ -18,7 +18,9 @@ store(storeName) value
 :   The name of the store to be defined.
 
 `value`
-:   An extended string value.
+:   An extended string value. This will be the initial default value.
+    Any persisted value stored previously with [`save()`](save) will
+    take precedence over `value`.
```
mcdurdin: This is not quite right. Changes to variable store values can only be seen when tested with `if()` -- variable stores should not be used with `any()` or other functions, and there is a function in the compiler to test that. Probably best to avoid confusing things...


```suggestion
:   An extended string value.

    If used to declare a [variable store](../guide/variable-stores), this 
    will be the initial default value. Any persisted value stored 
    previously with [`save()`](save) will take precedence over `value`. 
    Note that variable stores should not be used with [`any()`](any), 
    [`index()`](index), or [`outs()`](outs).
```

mcdurdin (dismissed):
Note that the variable-stores guide does clarify this also

developer/language/reference/store.md
ermshiperete: Thanks!

mcdurdin (approved):

PR #2522, opened 2026-05-27
fix: use ellipsis in stores.md example

### LINKED ISSUE #2517: bug: wrong rendering of `...` in markdown
If the markdown document contains code like [this](https://github.com/keymanapp/help.keyman.com/blob/master/developer/language/reference/save.md):

```
... > save(StoreName)
```

it gets [rendered as](https://help.keyman.com/developer/language/reference/save)

```
<code class="language-keyman"><span class="token operator">..</span>. <span class="token operator">&gt;</span> <span class="token rule-keyword keyword">save</span><span class="token punctuation">(</span>StoreName<span class="token punctuation">)</span></code>
```

(the `...` rendered as `<span class="token operator">..</span>.`).

Image

This looks wrong since the first two dots and the greater sign have slightly different colors than the third dot.

mcdurdin:
Also cleanup multi-line note in headers.md.

Note that there are over 1000 different usages of `...` throughout this site and I am not even going to attempt to look at them.

Fixes: #2517
Test-bot: skip


_includes/includes/index-content.php
```diff
@@ -7,6 +7,7 @@
  * be appended to the bottom of the list.
  */
 function get_file_titles_by_index($base, $unsorted_items) {
+  $items = [];
```
mcdurdin, line 10: fixes a background warning

ermshiperete (approved):

PR #2529, opened 2026-05-29
docs: add kb for ARM upgrade issues

rc-swag:
Upgrading Keyman for Windows from a non-arm supported version to a Arm support version does not work. This is possibly due to the way the keyman TIP dll is registerd by guid and the dll name. This kb explains the oldversion needs to be uninstalled before installing the newer version.

Test-bot: skip


knowledge-base/kb0120.md
```diff
@@ -0,0 +1,12 @@
+
+# Keyman for Windows with ARM Installation issues
```
mcdurdin:
```suggestion
# HOWTO: Upgrade Keyman for Windows 18.0 (or earlier) to 19.0+ on ARM devices
```

- no blank line
- doesn't really match the content precisely but the answer is there, and conveys information about the scenario in which the KB will be useful

mcdurdin (approved):

PR #2544, opened 2026-06-04
docs: add text tsf setting instructions

rc-swag:
Now that a second user has enqurired about the same issue it makes sense to bring the text from the forum into the kb so the steps are presserved and easy to link to for future quiries.

Test-bot: skip


Meng-Heng (approved):
LGTM, Ross!

PR #2548, opened 2026-06-07
feat: added support for Podman as container engine

Linsner:
This adds the Podmanfile for the help site to the project repo.

### DIFF
--- a/Podmanfile
+++ b/Podmanfile
@@ -1,0 +1,39 @@
+# syntax=docker/dockerfile:1
+
+ARG BUILDER_CONFIGURATION="release"
+FROM docker.io/library/php:7.4-apache@sha256:c9d7e608f73832673479770d66aacc8100011ec751d1905ff63fae3fe2e0ca6d AS composer-builder
+
+# Install Zip to use composer
+RUN apt-get update && apt-get install -y \
+    zlib1g-dev \
+    libzip-dev \
+    unzip
+RUN docker-php-ext-install zip
+
+# Install and update composer
+COPY --from=docker.io/library/composer /usr/bin/composer /usr/bin/composer
+RUN composer self-update
+
+WORKDIR /composer
+COPY composer.* /composer/
+# Consume the build argment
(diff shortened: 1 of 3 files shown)


Podmanfile
```diff
@@ -0,0 +1,39 @@
+# syntax=docker/dockerfile:1
+
```
mcdurdin, line 4:
```suggestion
# Please keep the content of this file in sync with Dockerfile.

```

And can we add the corresponding comment to Dockerfile also?

Podmanfile
```diff
@@ -0,0 +1,39 @@
...
+# Consume the build argment
+ARG BUILDER_CONFIGURATION
+RUN if [ "$BUILDER_CONFIGURATION" = "debug" ]; then \
+      # composer install --dev deprecated
+      COMPOSER_NO_DEV=0 composer install ; \
+    else \
+      COMPOSER_NO_DEV=1 composer install ; \
+    fi
+
+USER www-data
```
mcdurdin: This is above l.17 in `Dockerfile` (`WORKDIR /composer`). Is there a reason why it is here instead in `Podmanfile`?

mcdurdin (approved):
LGTM with the one small change to Dockerfile

mcdurdin:
Test-bot: skip

Podmanfile
Linsner: The USER instruction is actually not needed at all. I removed it and added explanatory comments about the differences.

Podmanfile
Linsner: Added the line in both files.

ermshiperete:
Would the `Podmanfile` work with Docker? I.e. do we need two separate files for Docker and Podman?

mcdurdin:
> Would the `Podmanfile` work with Docker? I.e. do we need two separate files for Docker and Podman?

See discussion in https://github.com/keymanapp/shared-sites/pull/96#discussion_r3371448309

